PR #26, opened 2022-11-29
Proof-of-concept `doc` command

stefanv:
Questions to resolve:

- Should we ensure that the correct requirements are installed beforehand? If so, how?
- Should we make SPHINXOPTS customizable? (Current is `-W`)
- Should we cater for directories other than `doc` (e.g., `docs`)?


devpy/cmds/_doc.py
```diff
@@ -0,0 +1,39 @@
...
+    if clean:
+        doc_dir = "./doc/build"
+        if os.path.isdir(doc_dir):
+            print(f"Removing `{doc_dir}`")
+            shutil.rmtree(doc_dir)
+
+    util.set_pythonpath(build_dir)
+
+    print("$ export SPHINXOPTS='-W'")
+    os.environ["SPHINXOPTS"] = "-W"
```
lagru, line 38: I think an option to override these could be useful. While iterating locally it can be useful to disable the `-W` flag, complete the build and clean up all the warnings afterwards. Otherwise the build fails on the first warning without showing the other ones.

jarrodmillman:
See #78 and #79.

PR #28, opened 2022-11-30
Also remove installation dir with `build --clean`

stefanv:
@rgommers Is your expectation that `./dev.py build --clean` should also erase the `build-install` directory?

My gut feel is yes, but I'm not 100% sure.


rgommers:
That makes sense to me, yes. I'll also note that I don't have any use for `--clean`, I use `git clean -xdf` which will be more reliable. The SciPy `dev.py` doesn't even have `--clean`, and we disabled the setuptools clean command ages ago because it was unreliable.

PR #34, opened 2022-12-11
Fix support for python 3.7

eli-schwartz:
As observed in https://github.com/qpv-research-group/solcore5/issues/238#issuecomment-1339626818

pyproject.toml lists support for 3.7 as a minimum. This is reasonable -- it will still be supported for another 6 months, and projects using devpy will frequently want to support that version too. As an infrastructure package, there's a strong rationale to be lenient in version support.

The snag is that 3.7 is not tested in CI, and it turns out it doesn't work. The practical minimum is 3.8 but both of the incompatibilities are trivially fixed -- and one of them is actually better code when using the fix.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,12 +16,15 @@ permissions:
 
 jobs:
   test_devpy:
+    strategy:
+      matrix:
+        python_version: ['3.7', '3.11']
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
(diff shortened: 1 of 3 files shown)


eli-schwartz:
ping?

eli-schwartz:
ping?

Abelarm:
Hey guys thank your for your great work, any news about the merging of this PR?

stefanv:
Sorry folks, I was away for a while, am back now.  In it goes!

eli-schwartz:
Thanks. :)

PR #35, opened 2022-12-19
Prefer site-packages paths with Python version

stefanv:
On Linux, you get python3.6/site-packages or python3/site-packages. This helps to ambiguate these paths for different versions of Python.

On Windows, it is typically \usr\Lib\site-packages, so we cannot rely on the version being present.

### DIFF
--- a/devpy/cmds/util.py
+++ b/devpy/cmds/util.py
@@ -36,10 +36,20 @@ def get_config():
 
 
 def get_site_packages(build_dir):
+    candidate_paths = []
     for root, dirs, files in os.walk(install_dir(build_dir)):
         for subdir in dirs:
             if subdir == "site-packages":
-                return os.path.abspath(os.path.join(root, subdir))
+                candidate_paths.append(os.path.abspath(os.path.join(root, subdir)))
+
+    if len(candidate_paths) == 0:
+        return None
+
+    # Prefer paths with current Python version in them
+    X, Y = sys.version_info.major, sys.version_info.minor
+    candidate_paths = sorted(candidate_paths, key=lambda p: f"python{X}.{Y}" in p)
+    candidate_paths = sorted(candidate_paths, key=lambda p: f"python{X}" in p)
+    return candidate_paths[-1]
 
 
 def set_pythonpath(build_dir):
--- a/devpy/cmds/util.py
+++ b/devpy/cmds/util.py
@@ -2,6 +2,7 @@ import os
 import sys
 import subprocess
 import shlex
+import sysconfig
 from pathlib import Path
 
 import click
@@ -39,25 +40,35 @@ def get_site_packages(build_dir):
     candidate_paths = []
     for root, dirs, files in os.walk(install_dir(build_dir)):
         for subdir in dirs:
-            if subdir == "site-packages":
+            if subdir == "site-packages" or subdir == "dist-packages":
                 candidate_paths.append(os.path.abspath(os.path.join(root, subdir)))
 
-    if len(candidate_paths) == 0:
-        return None
-
-    # Prefer paths with current Python version in them
     X, Y = sys.version_info.major, sys.version_info.minor
-    candidate_paths = sorted(candidate_paths, key=lambda p: f"python{X}.{Y}" in p)
-    candidate_paths = sorted(candidate_paths, key=lambda p: f"python{X}" in p)
-    return candidate_paths[-1]
 
+    site_packages = None
+    if any(f"python{X}." in p for p in candidate_paths):
+        # We have a system that uses `python3.X/site-packages` or `python3.X/dist-packages`
+        site_packages = [p for p in candidate_paths if f"python{X}.{Y}" in p]
+        if len(site_packages) == 0:
+            print(f"No site-packages found in `{build_dir}` for Python {X}.{Y}")
+            sys.exit(1)
+        else:
+            site_packages = site_packages[0]
+    else:
+        # A naming scheme that does not encode the Python major/minor version is used, so return
+        # the first site-packages path found
+        if len(candidate_paths) != 0:
+            site_packages = candidate_paths[0]
 
-def set_pythonpath(build_dir):
-    site_packages = get_site_packages(build_dir)
     if site_packages is None:
         print(f"No `site-packages` directory found under {build_dir}; aborting")
         sys.exit(1)
 
+    return site_packages
+
+
+def set_pythonpath(build_dir):
+    site_packages = get_site_packages(build_dir)
     env = os.environ
 
     if "PYTHONPATH" in env:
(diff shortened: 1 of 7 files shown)


mattip:
Is the version a concern? The c-extension modules are version-specific or abi3 which should be loaded by any version, right?

mattip:
Should there be a check that there is not more than one valid path?

stefanv:
@mattip The version is a concern if I build numpy for py311 and py310 side-by-side. What do you think is the logic for only one valid path?  Perhaps filter the list using `python3.10`, if that doesn't exist, filter by `python3`, if that doesn't exist just return whatever is available?

mattip:
> The version is a concern if I build numpy for py311 and py310 side-by-side

Why? What is the concern? The python files are identical, and the c-extensions have different names. I understand meson will create two directories on linux: `build-install/usr/lib/python3.10/site-packages` and `build-install/usr/lib/python3.11/site-packages`. One windows it will create a single `build-install/usr/Lib/site-packages`. So I think devpy should add the appropriate path to `sys.path`. There should never be two matches for the python under test.

mattip:
> Perhaps filter the list using python3.10, if that doesn't exist, filter by python3, if that doesn't exist just return whatever is available?

If that doesn't exist, error "something is wrong"

> in the case where there is only a python3.11 build and you are running on py310.

I would prefer a clear error when testing starts and not a "cannot import" error later on

stefanv:
But do we know what the path will look like on different platforms? Windows has no version info, Python seems to add major and sometimes minor version, not sure about MacOS?

mattip:
I think I would prefer this, does it make sense?:
```
def get_site_packages(build_dir):
    candidate_paths = []
    for root, dirs, files in os.walk(install_dir(build_dir)):
        for subdir in dirs:
            if subdir == "site-packages":
                candidate_paths.append(os.path.abspath(os.path.join(root, subdir)))
    if len(candidate_paths) == 0:
        raise ValueError(f"cannot find any 'site-packages' in {build_dir}")
    if len(candidate_paths) == 1:
        return candidate_paths[0]
    # Find paths with current Python version in them
    X, Y = sys.version_info.major, sys.version_info.minor
    major_paths = [p for p in candidate_paths if f"python{X}" in candidate_paths.split(os.sep)]
    if len(major_paths) > 1:
        raise ValueError(f"cannot find proper 'site-packages' to use in {build_dir}, more than one 'python{X}'")
    if len(major_paths == 1:
        return major_paths[0]
    minor_paths = [p for p in candidate_paths if f"python{X}.{Y}" in candidate_paths.split(os.sep)]
    if len(minor_paths) > 1:
        raise ValueError(f"cannot find proper 'site-packages' to use in {build_dir}, more than one 'python{X}.{Y}'")
    if len(minor_paths == 1:
        return minor_paths[0]
    raise ValueError(f"cannot find any 'site-packages' in {build_dir} suitable for python{X}.{Y}")
```

rgommers:
I recommend copying the logic from SciPy's `dev.py` exactly. It's reliable now AFAIK and was quite difficult to get right, there are lots of platform-specific variations.

I am seeing this problem locally as well now for some reason, where before it used to work. Not sure what changed, maybe Python version or some such thing.

stefanv:
> I recommend copying the logic from SciPy's `dev.py` exactly. It's reliable now AFAIK and was quite difficult to get right, there are lots of platform-specific variations.

That logic relies on `distutils`, and given this comment I didn't think that was the right approach:

```
# distutils is required to infer meson install path
# if this needs to be replaced for Python 3.12 support and there's no
# stdlib alternative, use CmdAction and the hack discussed in gh-16058
```

Directly searching for the path seems like a simpler alternative, if we can make it work?

> I am seeing this problem locally as well now for some reason, where before it used to work. Not sure what changed, maybe Python version or some such thing.

Perhaps you're on the previous version still? On `main` the path search has been reverted to its old behavior, and works the way it always did.

stefanv:
> I think I would prefer this, does it make sense?:

That looks correct, except that the more specific check (3.10) needs to come before the less specific check (3).

stefanv:
@mattip How about this version?

The exact heuristics that need to be replicated are the [ones from mesonpy](https://github.com/mesonbuild/meson-python/blob/main/mesonpy/__init__.py#L382), I suppose.

The logic here is simpler, because we can use the installed files to help guide us.

1. Search for any `site-packages` or `dist-packages`
2. If the results contain "python3.X", it means that the paths encode the major/minor version. In this case, ensure that our Python version is matched.
3. If not, we cannot distinguish between products for different versions of Python, so just return *any* site-packages path found.

devpy/cmds/util.py
```diff
@@ -36,18 +37,38 @@ def get_config():
...
+
+    X, Y = sys.version_info.major, sys.version_info.minor
+
+    site_packages = None
+    if any(f"python{X}." in p for p in candidate_paths):
+        # We have a system that uses `python3.X/site-packages` or `python3.X/dist-packages`
+        site_packages = [p for p in candidate_paths if f"python{X}.{Y}" in p]
+        if len(site_packages) == 0:
+            print(f"No site-packages found in `{build_dir}` for Python {X}.{Y}")
+            sys.exit(1)
```
mattip, line 54: It seems wrong for a utility function to call sys.exit. I would prefer it raise and the script runner catch the exception and exit

devpy/cmds/util.py
```diff
@@ -36,18 +37,38 @@ def get_config():
...
+        else:
+            site_packages = site_packages[0]
+    else:
+        # A naming scheme that does not encode the Python major/minor version is used, so return
+        # the first site-packages path found
+        if len(candidate_paths) != 0:
+            site_packages = candidate_paths[0]
```
mattip, line 61: This should check that only one such directory exists

mattip:
No tests?

stefanv:
I'll add tests and see if there's a convenient way to change exiting to exceptions, thanks.

mattip:
Any progress?

stefanv:
Sorry, I was a bit pre-occupied with the annual festive transition; back in the saddle now!

devpy/tests/test_util.py
```diff
@@ -1,10 +1,84 @@
...
+        install_dir = pjoin(d, "build-install")
+        make_paths(
+            install_dir,
+            [
+                f"/usr/lib64/python{X}.{Y}/site-packages/",
+                f"/usr/lib64/python{X}.{Y + 1}/site-packages/",
+                f"/usr/lib64/python{X}.{Y + 2}/site-packages/",
+            ],
+        )
+        util.get_site_packages(build_dir)
```
mattip, line 33: Should this check that the returned value is the one expected?

mattip:
Thanks. Looks good, maybe a bit of test improvement possible but fine to do that in subsequent PRs. As this matures, it would be good to add a version and put it on PyPI

stefanv:
I added a test for return values. I will add testing on Windows in another PR; it may require a bit of tweaking to get right.

mattip:
Thanks

PR #36, opened 2023-01-11
Test on Windows and MacOS as well

stefanv:


stefanv:
Implemented on `main`.

Issue #38, opened 2023-01-17
Additional windows problems


mattip:
From the meson branch of NumPy, using the HEAD of this repo 2ad0423b8 in a virtualenv of the windows store CPython3.9 located in `d:'\temp\cpython_39`:
- When starting from a cmd shell without cl on the path, `python dev.py build` fails to set up MSVC in a way that ninja can call it. I need to do this (something kindly prints out the second line so I know what to call, then the second invocation of dev.py installs to `build-install`). A work-around is to activate MSVC before calling this command, but I think it should Just Work without that.
  ```
  python dev.py build
  D:\temp\cpython_39\Scripts\meson compile -C build
  python dev.py build
  ```

- running `python dev.py test` still fails. It seems to be calling the base-python rather than the one inside the virtualenv. We touched on this in [the other issue](https://github.com/scientific-python/devpy/issues/32#issuecomment-1379688740). If I change [this line](https://github.com/scientific-python/devpy/blob/2ad0423b8a7b1eb7f23192f07f7743a660ab71ba/devpy/cmds/_test.py#L35) to use `sys.executable` instead of `"python"` then it works.

eli-schwartz:
> * When starting from a cmd shell without cl on the path, `python dev.py build` fails to set up MSVC in a way that ninja can call it. I need to do this (something kindly prints out the second line so I know what to call

Meson's Ninja backend can print out that message for you:
```python
if self.build.need_vsenv:
    mlog.log('Visual Studio environment is needed to run Ninja. It is recommended to use Meson wrapper:')
```

Running `meson compile` prints this out:
```python
vsenv_active = setup_vsenv(b.need_vsenv)
if vsenv_active:
    mlog.log(mlog.green('INFO:'), 'automatically activated MSVC compiler environment')
```

Sadly, it's impossible set up VS permanently from Meson. We can detect it even not on PATH, but we can only set it up in child processes of Meson itself (that's why `meson compile` exists, mostly).

This is really a flaw in ninja, there's a ticket at https://github.com/ninja-build/ninja/issues/1610#issuecomment-1377743206 discussing it.

stefanv:
@mattip Do you have a Windows machine to test on? If so, do you think you can make a PR? Looks like it should be a very small change.

stefanv:
Also, I test suite shouldn't be passing on Windows, sounds like?

mattip:
[...]

PR #41, opened 2023-01-27
Remove Strawberry GCC compiler in Windows CI

stefanv:


.github/workflows/test.yml
```diff
@@ -44,6 +44,13 @@ jobs:
           cd example_pkg
           ./dev.py build
           ./dev.py test
+      - name: Patch Windows runner
+        if: matrix.os == 'windows-latest'
+        run: |
+          Remove-Item -Path C:\Strawberry -Recurse
+          Remove-Item -Path C:\ProgramData\Chocolatey\bin\gcc.exe
+          Remove-Item -Path C:\msys64 -Recurse -Force
+          Remove-Item -Path "C:\Program Files\LLVM" -Recurse
```
mattip, line 53: How would this work when using devpy in NumPy? Would we need to add this CI step?

stefanv: I guess it depends which compiler you want. If you're happy with clang, presumably you can just remove gcc. These steps are only necessary to simulate having the Visual Studio compilers.

eli-schwartz: Yup -- this isn't really a devpy thing anyway. It happens for random C/C++ projects that don't have anything to do with python, also.

Using something like https://github.com/ilammy/msvc-dev-cmd is a good idea IMO, at least for GitHub actions. It guarantees that `cl.exe` is set up correctly *and* is somewhere in PATH, and if it's in PATH then it will always win as far as Meson is concerned.

eli-schwartz: This PR now correctly simulates an identical setup in CI to compare to the case where an end user on a personal laptop has MSVC installed, but never went and installed some other gcc/clang compiler.

i.e. it simulates the common case.

mattip: :+1:

stefanv:
rf #38

Issue #40, opened 2023-01-26
Test command behavior needed for NumPy's use of devpy


rgommers:
@stefanv as a follow up to our chat today and to gh-13 (which covered multiple things), here are the issues with the `test` command I ran into for NumPy:

- The default for NumPy should be to run the `'fast'` tests, not `'full'`. I.e, the equivalent of `np.test()`.
- The `test` command on NumPy should not invoke `pytest` directly, but invoke `PytestTester` from `numpy/_pytesttester.py` - that will solve a bunch of things, including the `fast` mode mentioned above.
- Being able to run per-submodule tests is very useful. And a few other things as well that SciPy's `dev.py` has. All the following apply for NumPy unchanged (I left out `--coverage`, that's too niche I'd say):
```
--verbose    -v             more verbosity    
--parallel   -j  N_JOBS     Number of parallel jobs for testing            
--durations  -d  NUM_TESTS  Show timing for the given number of slowest tests 

--submodule  -s  MODULE_NAME  Submodule whose tests to run (cluster, constants, ...) 
--tests      -t  TESTS        Specify tests to run
--mode       -m  MODE         'fast', 'full', or something that could be passed to pytest -m
                              as a marker expression   
                              [default: fast]
```

- It's also quite useful to get the `test` command to trigger the `build` command by default, to avoid having to do so manually or running the tests without picking up the latest local changes a user has.

mattip:
One more thing I noticed when trying to run the tests under a debugger: since devpy starts a new process, I cannot do `gdb --args python dev.py test ...`. If this is a requirement of the way `devpy` loads numpy from `build-install/...`, there should be a way to add a debugging mode.

mattip:
> It's also quite useful to get the test command to trigger the build command by default

meson-python has support for [editable installs](https://meson-python.readthedocs.io/en/latest/how-to-guides/editable-installs.html#editable-installs), maybe devpy could take something from there?

rgommers:
Editable installs are still in early testing, and we still have to wait and see whether they get on par with the approach in devpy.

WarrenWeckesser:
In the meantime, here are some notes on pytest arguments that can be used to fine-tune `spin test` and replicate some of the options in `dev.py`.

* To use more than one CPU to run the tests:

      spin test -- -n NUM_CPUS

  E.g.

      spin test -- -n 8

  For more details, see https://pytest-xdist.readthedocs.io/en/latest/distribution.html

* To run the tests for a specific submodule:

[...]

PR #42, opened 2023-01-30
DOC: add some hints to build, test

mattip:
Some things I found helpful when trying this out
- document that the tests will run "full" by default, and show how to skip slow tests
- document adding CFLAGS for debugging. The meson FAQ seemed to suggest this is best practices, maybe there is a better way?

### DIFF
--- a/devpy/cmds/_build.py
+++ b/devpy/cmds/_build.py
@@ -18,12 +18,16 @@ from .util import run, install_dir
 def build(build_dir, meson_args, jobs=None, clean=False, verbose=False):
     """🔧 Build package with Meson/ninja and install
 
-    MESON_ARGS are passed through directly to pytest, e.g.:
+    MESON_ARGS are passed through e.g.:
 
     ./dev.py build -- -Dpkg_config_path=/lib64/pkgconfig
 
     The package is installed to BUILD_DIR-install
 
+    By default builds for release, to be able to use a debugger set CFLAGS
+    appropriately. For example, for linux use
+
+    CFLAGS="-O0 -g" ./dev.py build
     """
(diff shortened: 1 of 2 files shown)


eli-schwartz:
> document adding CFLAGS for debugging. The meson FAQ seemed to suggest this is best practices, maybe there is a better way?

These particular flags are really equivalent to passing the meson_args `-Dbuildtype=debug`.

mattip:
Does `-Dbuildtype=debug` require a debug build of python either on linux or windows?

mattip:
The first point is connected to #40

eli-schwartz:
> Does `-Dbuildtype=debug` require a debug build of python either on linux or windows?

Definitely not on Linux, since it maps to exactly the combination of `-Doptimization=0 -Ddebug=true`.

Good point about Windows, since by default the value of `b_vscrt` is `from_buildtype` and per the table at https://mesonbuild.com/Builtin-options.html this will map to also setting `/MDd`.

(I have no idea whether it is compatible with a release build of python itself.)

mattip:
The problem with `/MDd` is that it will link to a different runtime library, so it will require a debug build of python for windows. Then this will percolate out to all other c-extension modules, so you must rebuild the entire environment. 

I usually set the cflags "/Od /Zi" instead, which is much more like "-O0 -g" for gcc. Well, that is what I tell people I do. In reality, since using a debugger in windows is painful, I first use a lot of debug prints to see if I can avoid using a debugger at all, so then all I care about is having a small failing test and a quick build turn-around.

stefanv:
Thanks for this, Matti!

PR #43, opened 2023-02-02
Fix Windows builds: use `meson compile` instead of `ninja`

### LINKED ISSUE #38: Additional windows problems

stefanv:
Closes #38

### DIFF
--- a/devpy/cmds/_build.py
+++ b/devpy/cmds/_build.py
@@ -50,7 +50,7 @@ def build(build_dir, meson_args, jobs=None, clean=False, verbose=False):
         shutil.rmtree(build_dir)
         run(build_cmd)
 
-    run(["ninja", "-C", build_dir])
+    run(["meson", "compile", "-C", build_dir])
     run(
         [
             "meson",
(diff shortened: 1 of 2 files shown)


mattip:
Seems to work, thanks!

PR #49, opened 2023-02-27
Add installation instructions

### LINKED ISSUE #46: Installing devpy
Would it be possible to add installation instructions to the README? Currently doing ``pip install devpy`` appears to install an unrelated package, so even if this tool is not yet on PyPI it would be worth telling people to not install devpy from PyPI?

stefanv:
Closes #46


stefanv:
@astrofrog is this what you had in mind?

astrofrog:
Yes looks great!

Issue #50, opened 2023-02-28
Run build command before tests


stefanv:
See https://click.palletsprojects.com/en/8.1.x/advanced/#invoking-other-commands

We don't need to pass through any option; this is mainly a safety check to ensure that tests aren't run against a modified version.

Note that Meson will not be the default build backend for every project, so for now just look at the `build` cmd defined in the tool configuration, and run that.

PR #51, opened 2023-03-01
Run `build` before `test` command

stefanv:
In order to run the `build` command reliably without arguments, the `--build-dir` flag was removed from all commands.

Closes gh-50

### DIFF
--- a/devpy/cmds/_build.py
+++ b/devpy/cmds/_build.py
@@ -6,16 +6,13 @@ from .util import run, install_dir
 
 
 @click.command("build")
-@click.option(
-    "--build-dir", default="build", help="Build directory; default is `$PWD/build`"
-)
 @click.option("-j", "--jobs", help="Number of parallel tasks to launch", type=int)
 @click.option("--clean", is_flag=True, help="Clean build directory before build")
 @click.option(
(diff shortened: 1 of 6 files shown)


devpy/cmds/_build.py
```diff
@@ -29,18 +26,17 @@ def build_meson(build_dir, meson_args, jobs=None, clean=False, verbose=False):
 
     CFLAGS="-O0 -g" ./dev.py build
     """
-    build_dir = os.path.abspath(build_dir)
-    inst_dir = install_dir(build_dir)
+    build_dir = os.path.abspath("build")
```
ganesh-k13, line 29: Quick Question: Does this mean we loose the flexibility to use different names for the build directory? I think for Meson it's OK as it defaults to `"build"` but other build systems in the future may not?

stefanv: I think the next refactoring I'll do will separate meson build & test from `build`-based build & test, with perhaps other options to come. The meson test process is fundamentally different from how you would approach it with installed wheels or editable installs. Eventually, they may come together again, when Meson's editable install is properly fleshed out, we'll see.

PR #52, opened 2023-03-02
Group meson commands in preparation for other build systems

stefanv:
This refactoring makes it easier to add commands for other build systems.
Backward compatibility is preserved for now so as to not break existing projects.


README.md
```diff
@@ -45,12 +81,22 @@ On Unix-like systems, you can also copy the [`dev.py` script](https://github.com
 
 ## Built-in commands
 
+### [Meson](https://meson-python.readthedocs.io)
```
lithomas1, line 84: https://mesonbuild.com/ should be the right link here.

devpy does not invoke meson-python through the devpy meson commands so this could be confusing.

devpy/cmds/__init__.py
```diff
@@ -1,3 +1,8 @@
-from ._build import build
-from ._test import test
-from ._shell import ipython, python, shell
+from . import meson
+
+# Backward compatibility with older versions
+build = meson.build
```
lithomas1, line 4: Can we issue a deprecation warning for people still using these directly?

README.md
stefanv: The build backend is `mesonpy`, but the build engine is `meson`. My thought was that this is an easier place to start:

https://meson-python.readthedocs.io/en/latest/tutorials/introduction.html#tutorial-introduction

devpy/cmds/__init__.py
stefanv: Sure, do you know of an easy way to implement that?

lithomas1: You can try a module-level __getattr__ (for python 3.7+), that raises the Deprecation/FutureWarning.

PR #63, opened 2023-03-15
Rename package

jarrodmillman:
https://pypi.org/project/spin/


CHANGELOG.md
```diff
@@ -1,72 +1,72 @@
 # Changelog
 
-## [v0.1](https://github.com/scientific-python/devpy/tree/v0.1)
+## [v0.1](https://github.com/scientific-python/spin/tree/v0.1)
```
stefanv, line 3: Should we leave this as-is, as only update upcoming releases?

pyproject.toml
```diff
@@ -1,11 +1,11 @@
 [project]
-name = "devpy"
-version = "0.2rc1.dev0"
+name = "spin"
+version = "0.2b1"
```
stefanv, line 3: Just for my own education, what's the beta1 about?

stefanv (commented):
Looks good, I will make a minor push to the PR with a README update as well.

pyproject.toml
jarrodmillman: Whoops, not sure why I left that there...

stefanv (approved):

PR #65, opened 2023-03-20
Add back PY extension to example stub

lagru:
While stripping the PY extension saves on typing, it hides the file from tools that rely on the extension to identify Python files. E.g projects linting with `ruff check .` will not notice that their `spin` script isn't checked.

Shortening the name may not be worth deviating from the established assumption that Python files end with `.py` which is also followed by similar tools such as NumPy's `runtest.py`.

Prompted by this discussion in https://github.com/scikit-image/scikit-image/pull/6842#discussion_r1140102777.


lagru:
I noticed that `grep -R 'dev.py'` still picks up a few references to the old name that I don't think are intended? I'm happy to fix those here or in another PR depending on how this suggestion is received. :)

```
spin/cmds/meson.py:    ./dev.py build -- -Dpkg_config_path=/lib64/pkgconfig
spin/cmds/meson.py:    CFLAGS="-O0 -g" ./dev.py build
spin/cmds/meson.py:                "Meson configuration failed; please try `dev.py build` again with the `--clean` flag."
spin/cmds/meson.py:    ./dev.py test -- -v
spin/cmds/meson.py:    ./dev.py test -- -m "not slow"
spin/cmds/meson.py:    ./dev.py ipython -- -i myscript.py
spin/cmds/meson.py:    ./dev.py shell -- -c 'echo $PYTHONPATH'
spin/cmds/meson.py:    ./dev.py python -- -c 'import sys; print(sys.path)'
README.md:On Unix-like systems, you can also copy the [`dev.py` script](https://github.com/scientific-python/spin/blob/main/example_pkg/dev.py) to the root of your project directory, and launch it as:
README.md:./dev.py
README.md:The `dev.py` tool was [proposed for SciPy](https://github.com/scipy/scipy/issues/15489) by Ralf Gommers and [implemented](https://github.com/scipy/scipy/pull/15959) by Sayantika Banik, Eduardo Naufel Schettino, and Ralf Gommers (also see [Sayantika's blog post](https://labs.quansight.org/blog/the-evolution-of-the-scipy-developer-cli)).
```

stefanv:
Thanks, Lars. I'd prefer to leave the name as `spin` here. The file is three lines long, so the linting justification feels weak. Is there another reason why we should rename it? E.g., can you show that it doesn't work on Windows without the `.py` extension?

A PR fixing the other occurrences of `dev.py` would be welcome!

lagru:
I like the route to provide the `spin` command as an installed script! And in that case I'm totally fine with dropping the .py extension. So closing this. 

It seems like the old references to `dev.py` were already cleaned up in the meantime.

PR #70, opened 2023-03-29
Add meson.run command

### LINKED ISSUE #69: Run a command in the shell without invoking the shell
Hi,

I'm currently using meson-python to build a package locally and also do development. However, beyond just running unit-tests, I would also like to run certain Python files (e.g. example scripts) and test them out locally. 

The current workflow I have is:

1. build and install `Dev.py build -j 4`
2. launch a shell `dev.py shell`
3. add the build to the PATH
4. Run Python stuff inside

This is kinda tedious because it just involves a bunch of commands occurring in different shells technically. I was wondering if it is possible to implement something like:

`./dev.py run <shell command>`

E.g. `./dev.py run ./example.py` which runs `example.py` file inside the shell.

Note this is very similar to what for example poetry has with `poetry run <shell command>`.

stefanv:
The command operates in two modes. If a single argument is given, it uses subprocess's `shell=True`, so you can use shell substitution, as in:

  spin run 'echo $PYTHONPATH'

If multiple arguments are given, it uses standard execution, e.g.:

  spin run python -c 'import sys; del sys.path[0]; import example_pkg'

Enabling `shell=True` for the latter would incorrectly disregarded the quoting.

Closes #69

### DIFF
--- a/.github/workflows/test.sh
+++ b/.github/workflows/test.sh
@@ -12,6 +12,9 @@ prun spin docs
 
 SPIN_PYTHONPATH=$(spin run 'echo $PYTHONPATH')
 echo spin sees PYTHONPATH=\"${SPIN_PYTHONPATH}\"
-[[ $SPIN_PYTHONPATH == *"site-packages" ]]
+$ if [[ ${SPIN_PYTHONPATH} == "\$PYTHONPATH" ]]; then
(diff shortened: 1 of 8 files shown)


stefanv:
I can't quite get Windows to do what I want, but I'm getting hold of a physical test machine soon which should help debugging significantly.

jarrodmillman (approved):

stefanv:
Phew :sweat_smile:

PR #73, opened 2023-06-08
Test on Python 3.12.0-beta.2

jarrodmillman:


stefanv:
See https://github.com/python/cpython/issues/105524

PR #81, opened 2023-07-01
Use ruff

jarrodmillman:


stefanv:
Can we hold off until the run cmd is merged? Then happy to put this in.

example_pkg/.spin/cmds.py
```diff
@@ -14,7 +16,7 @@ def example(flag):
     click.secho("Running example custom command", bold=True, fg="bright_blue")
     print()
     config = util.get_config()
-    commands = util.get_commands()
```
stefanv, line 17: This is for illustrative purposes, so would like to leave it in. Maybe I should just print the commands out.

stefanv (approved):
Looks great! Let me just fix up that the `get_commands`.

PR #86, opened 2023-07-01
Use changelist

jarrodmillman:


RELEASE.md
```diff
@@ -14,8 +14,9 @@ Example `version`
 
 - Update and review `CHANGELOG.md`:
 
-      gem install github_changelog_generator
-      github_changelog_generator -u scientific-python -p spin --since-tag=<last tag>
+      changelist scientific-python/spin  <v0.0> main --version <0.1> >> CHANGELOG.md
```
stefanv, line 17:
```suggestion
      changelist scientific-python/spin v0.0 main --version v0.1 > CHANGELOG.md
```

Not sure if that should be v0.1 or 0.1

RELEASE.md
```diff
@@ -14,8 +14,9 @@ Example `version`
...
+
+  where <v0.0> is the last release and <0.1> is the new one.
```
stefanv, line 19:
```suggestion
  replacing v0.0 with the previous release and v0.1 is the new one.
```

Not sure if that should be `v0.1` or `0.1`?

stefanv (approved):

RELEASE.md
jarrodmillman: We can take out the angle brackets indicating what needs to be changed. But we need to leave in the `>>`.

RELEASE.md
jarrodmillman: The `0.1` is used in the CHANGELOG and the `v0.0` is a tag. We've been prepending `v` to the release tags, which is pretty common and we should stick with it. I think the release number shouldn't have the `v` prepended. If it does, then the CHANGELOG will say things like:
```
We're happy to announce the release of spin v0.1!
```

I think it looks better the way I had it
```
We're happy to announce the release of spin 0.1!
```

Feel free to make a PR to make whatever changes you want. I don't have a strong opinion on this. I will copy your changes to the other sites to keep things consistent.

RELEASE.md
jarrodmillman: If you take out the angle brackets, we should make it clear that that text needs to be changed in another way. Maybe just explain it in the text below.

jarrodmillman: An argument for keeping the angle brackets is that is the standard notation for indicating a manadatory argument that needs to be replaced when using.

jarrodmillman: Maybe you would prefer
```
      changelist scientific-python/spin  <tag of the previous release> main --version <upcoming release version> >> CHANGELOG.md
```

stefanv: It'd be nice to be able to copy, paste, and modify. The brackets are a bit noisy and hard to read.

PR #92, opened 2023-07-17
Suggest setting meson buildtype for a debug build

ngoldbaum:
This suggestion will be correct on all platforms meson supports.


stefanv:
Thank you!

PR #93, opened 2023-07-17
DEP: migrate from toml package to tomllib

neutrinoceros:
The `toml` package is unmaintained. `tomli` is the standard replacement and is available in the standard library since Python 3.11 as `tomllib`.

See https://peps.python.org/pep-0680/


stefanv:
Excellent, thank you!

PR #97, opened 2023-08-11
Allow custom Meson CLI path

stefanv:
@rgommers This PR allows setting the Meson CLI in `pyproject.toml`. Do you think that would address NumPy's need?


rgommers:
Close I think. `cli = 'path/to/custom/meson'` still will not work on Windows, so I think we need a convention here that if `cli` ends in `.py`, then `spin will prepend `sys.executable` to the command.

stefanv:
Thanks, done.

jarrodmillman (approved):

rgommers:
This looks good now - thanks for the release. I'll try this out later this week in `numpy`.

PR #107, opened 2023-09-12
Throughout, invoke build before commands

stefanv:
May need a bit of additional work to suppress build output from appearing in stdout, if it is desirable do do, e.g.:

```
FOO=$(spin run something)
```

Steps:
- Use contextlib redirect_stdout around _run cmd
- Extend build to allow a quiet mode, in which `output=False` on each `_run` call


jarrodmillman (approved):

stefanv:
Hrm that CI failure wasn't triggered by the code changes; something with Ubuntu update :/

PR #124, opened 2023-09-28
Also support build sanity check on <3.11

stefanv:
/cc @mattip


mattip:
Thanks, LGTM. Maybe add a comment that pytest swallows the chained error, apparently since it occurs when importing conftest.py very early in pytest startup

jarrodmillman (approved):

PR #129, opened 2023-10-25
Support .spin.toml/spin.toml as an alternate configuration files

### LINKED ISSUE #128: Support alternate configuration file .spin.toml
I would like to propose support for alternate configuration file names, let say `.spin.toml`, that would take precedence over `pyproject.toml` if present in the source tree. 

Would you guys accept a PR implementing this feature?

Rationale:

* Support using `spin` in projects that may not be strictly related to Python and where `pyproject.toml` would look awkward/confusing to have in the source tree.
* Prevent clutter in `pyproject.toml`. Some developers prefer to keep `pyproject.toml` as minimal as possible. So let's be open-minded and let users decide on base to their preferences.
* Follow other Python tools (`pytest`, `ruff`, `mypy`), even if they may be doing it for legacy reasons.

dalcinl:
Fixes #128.

### DIFF
--- a/spin/__main__.py
+++ b/spin/__main__.py
@@ -1,3 +1,4 @@
+import collections
 import importlib
 import importlib.util
 import os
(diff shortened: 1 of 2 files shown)


stefanv:
Thanks @dalcinl! I pushed a commit to simplify the config loader logic a bit; take a look and let me know what you think.

dalcinl:
Your changes changed the behavior of what I did. 

My code had support for a `.spin.toml` file with the following contents (note the `tool.spin` section is implicit in this case).

```toml
package = 'mypkg'
[commands]
"Build" = [
  "spin.cmds.meson.build",
  "spin.cmds.meson.test",
]
```

stefanv:
Yes, I changed it to make the documentation simpler, but if you feel strongly about it we can change it back.

dalcinl:
> Yes, I changed it to make the documentation simpler, but if you feel strongly about it we can change it back.

Well, it is not really about feelings, but rather convenience to users, which should not trump over documentation difficulties. Again, I'm only following common practice from other projects.

> But you are also missing a section: the project name.

I'm not sure what do you meant. Nonetheless, I found a flaw in my previous logic that may be related to your comment. 

I'll try my best to rework the code, but it will still be a bit convoluted. If you still do not like it, I'll surrender.

PS: Maybe everything would get easier if, rather than erroring, we would support a hierarchical configuration using `collections.ChainMap` to honor `.spin.toml` first (and maybe other names like `spin.toml` and `.spin/config.toml` in a documented order), with a fallback in `pyproject.toml`. What do you think?

stefanv:
If it simplifies the logic, I'm fine with that. Thanks @dalcinl for working out the details.

dalcinl:
@stefanv Please take a fresh new look. The most important changes with respect to my previous approach are:

* Added support for `spin.toml` without initial dot `.`, for those that do not like hidden files.
* Configuration is now hierarchical and resolve in the order `.spin.toml`, `spin.toml`, `pyproject.toml`
* I gave up on supporting an implicit `[tool.spin]` section in `.spin.toml`.
* I explicitly error if `tool.spin.commands` is not present. IMHO, that's better than a traceback with KeyError.

Feel free to push on top any change you may consider necessary, and thanks you for taking the time to review my previous mess.

spin/__main__.py
```diff
@@ -19,35 +20,53 @@
...
+        if not os.path.exists(filename):
+            return None
+        with open(filename, "rb") as f:
+            try:
+                return tomllib.load(f)
+            except tomllib.TOMLDecodeError:
+                error("cannot parse [{filename}]")
+
+    toml_config = collections.ChainMap()
+    toml_config.maps.extend(
```
stefanv, line 37: Any reason not to instantiate `ChainMap` directly from these dicts? We could have `load_toml` above return `{}` instead of None.

stefanv (commented):
This is looking good to me! I left one minor comment.

spin/__main__.py
dalcinl: Well, I wanted to add the loaded configurations as children of the ChainMap, in case in the future we need to store things manually on it (that would go to the (currently empty) first dict in the chain (`maps` attribute)). Regarding the returning `None`, I wanted to preserve that return value as a signal that the file is not present. Anyway, the change `None` -> `{}` is inconsequential for the rest of the code, as an empty dict will not be appended to the chain.

PS: Just in case I was not clear enough, a ChainMap created with no args is born with an initial empty dict in the chain. Appending to the `maps` list adds new dicts to the chain.

stefanv (approved):

spin/__main__.py
stefanv: Makes sense; I'm not sure we'd ever need to distinguish between manual parameters and the rest, but at this point we're in the weeds, so in it goes.

dalcinl: > I'm not sure we'd ever need to distinguish between manual parameters and the rest

Think about third parties writing their own commands wrapping other commands and wanting to mess around with config to hack/monkeypatch things.

stefanv:
Thanks @dalcinl!

PR #137, opened 2023-11-08
Add LLDB (debug) command

### LINKED ISSUE #125: Create a `spin lldb` similar to `spin gdb`.
Using `gdb` on a mac is an exercise in pain because of SIP and code signing. People will have a much easier time using `lldb` so if possible we should create a new `spin` subcommand for it.

stefanv:
LLDB is the default debugger in Xcode on macOS, but also runs on Linux and Windows. See https://lldb.llvm.org/

Closes #125

/cc @ngoldbaum

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,8 @@ python -m spin
 
 ### [Meson](https://meson-python.readthedocs.io)
 
+Available as `spin.cmds.meson.*`.
+
 ```
   build    🔧 Build package with Meson/ninja and install to `build-install`
   ipython  💻 Launch IPython shell with PYTHONPATH set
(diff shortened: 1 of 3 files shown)


ngoldbaum:
Ping @mtsokol who has a debugger working with numpy on a mac and can test this on MacOS.

mtsokol:
I've just run `spin lldb` with locally built spin and it worked same as with `lldb` directly.

stefanv:
Thanks for the feedback @mtsokol. Does the spin command help simplify your workflow?

mtsokol:
Yes, I think it's more consistent now to use commands `spin test`, `spin ipython` and `spin lldb` for me.

jarrodmillman (approved):

Issue #77, opened 2023-06-30
Generalize path setting mechanism to cover various build backends (meson, build, pip, etc.)


stefanv:
At the moment, we only have support for Meson. But, really, the only thing that would distinguish a meson command from a pip/build one is the way we set PYTHONPATH.

Generalize the path setting functionality, so that we can use the same commands on all those systems.

stefanv:
I don't have any good ideas on how to solve this atm.

PR #141, opened 2023-11-14
POC: generic commands

stefanv:
I'm not super happy with the complexity this PR introduces, but it does demonstrate how to implement commands that work with meson and with editable pip installations.

This PR does not allow them to work with both at the same time; perhaps that is desirable.

This is a first attempt to think more about #77


stefanv:
See #155

PR #146, opened 2024-01-04
Add `gcov` flags for build

ganesh-k13:
### Changes
- Added `gcov` flags for build

### Related
This is needed to generate `gcov` reports for NumPy. xref: https://github.com/numpy/numpy/pull/24992

### Testing
In NumPy, after merging of https://github.com/numpy/numpy/pull/24992:

image

1. `spin test --coverage --gcov`
2. `spin test --generate-lcov-html`
```
~/os/numpy (bld_24080_coverage) » spin test --generate-lcov-html                                                                                                            ganesh@ganesh-MS-7B86
Deleting old HTML coverage report...
Generating HTML coverage report...
$ ninja coverage-html -C build
Coverage report generated successfully and written to /home/ganesh/os/numpy/build/meson-logs/coveragereport/index.html
```

---

Other error handling
1. `spin build`
2. `spin test --generate-lcov-html`
```
~/os/numpy (bld_24080_coverage) » spin test --generate-lcov-html                                                                                                            ganesh@ganesh-MS-7B86
Error: GCOV files missing... Cannot generate coverage reports. Coverage reports can be generated by `spin test --coverage --gcov`

```

---

### Notes
Although I verified it with NumPy, ~needs a bit more testing which I'll do in a while~
```
~/os/numpy (bld_24080_coverage*) » find . -name \*.gcno | wc -l                              ganesh@ganesh-MS-7B86
204
```

### DIFF
--- a/.github/workflows/test.sh
+++ b/.github/workflows/test.sh
@@ -3,11 +3,13 @@ set -e
 PLATFORM=$(python -c 'import sys; print(sys.platform)')
 
 RED="\033[31;1m"
+BLUE="\033[34m"
 MAGENTA="\033[35m"
 NORMAL="\033[0m"
 
 ptest() { echo -e "\n${MAGENTA}[TEST] $@${NORMAL}\n" ; }
-prun() { echo -e "$RED\$ $@ $NORMAL" ; "$@" ; }
+perror() { echo -e "\n${RED}[ERROR] $@${NORMAL}\n" ; }
+prun() { echo -e "${BLUE}\$ $@ ${NORMAL}" ; "$@" ; }
 
 prun cd example_pkg
 
@@ -19,19 +21,19 @@ pip install meson-python ninja
 
 
 # Test spin build + debug builds
-echo -e "${MAGENTA}Creating debug builds${NORMAL}"
+echo "Creating debug builds"
 prun spin build --gcov
-echo -e "${MAGENTA}Did the build folder get generated?${NORMAL}"
+ptest Did the build folder get generated?
 if [ ! -d "build" ] || [ ! -d "build-install" ]; then
-    echo -e "${RED}build and/or build-install folders did not get generated${NORMAL}"
+    perror build and/or build-install folders did not get generated
     exit 1
 else
     echo "Yes"
 fi
-echo -e "${MAGENTA}Does the debug build contain gcov files?${NORMAL}"
+ptest Does the debug build contain gcov files?
 matching_files=$(find . -type f -name "*.gc*")
 if [ -z "$matching_files" ]; then
-    echo -e "${RED}Debug files did not get generated${NORMAL}"
+    perror Debug files did not get generated
     exit 1
 else
     echo "Yes"
@@ -41,7 +43,7 @@ ptest Does spin expand \$PYTHONPATH?
 SPIN_PYTHONPATH=$(spin run 'echo $PYTHONPATH')
 echo spin sees PYTHONPATH=\"${SPIN_PYTHONPATH}\"
 if [[ ${SPIN_PYTHONPATH} == "\$PYTHONPATH" ]]; then
-    echo "Expected Python path, but got $SPIN_PYTHONPATH instead"
+    perror Expected Python path, but got $SPIN_PYTHONPATH instead
     exit 1
 fi
 
@@ -58,7 +60,7 @@ VERSION=$(spin run python -c 'import sys; del sys.path[0]; import example_pkg; p
 if [[ $VERSION == "0.0.0dev0" ]]; then
     echo "Yes"
 else
-    echo "No, output is $VERSION"
+    perror No, output is $VERSION
     exit 1
 fi
 
@@ -68,7 +70,7 @@ OUT=$(spin run ls)
 if [[ $OUT == *"Warning! An editable installation"* ]]; then
     echo "Yes"
 else
-    echo "No"
+    perror No
     exit 1
 fi
 prun pip uninstall --quiet -y example_pkg
@@ -82,7 +84,7 @@ if [[ $PLATFORM == linux || $PLATFORM == darwin ]]; then
     if [[ $OUT == *"Did you mean to call"* ]]; then
         echo "Yes"
     else
-        echo "No, output is: $OUT"
+        perror No, output is: $OUT
         exit 1
     fi
 fi
--- a/.github/workflows/test.sh
+++ b/.github/workflows/test.sh
@@ -16,7 +16,26 @@ prun spin --version
 
 ptest build command runs
 pip install meson-python ninja
-prun spin build
+
+
+# Test spin build + debug builds
+echo -e "${MAGENTA}Creating debug builds${NORMAL}"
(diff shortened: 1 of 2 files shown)


spin/cmds/meson.py
```diff
@@ -130,6 +137,18 @@ def build(meson_args, jobs=None, clean=False, verbose=False, quiet=False):
     """
     build_dir = "build"
     setup_cmd = _meson_cli() + ["setup", build_dir, "--prefix=/usr"] + list(meson_args)
+    env = os.environ
+
+    if gcov_build:
+        # Ref: https://github.com/numpy/numpy/pull/3979/files
```
ganesh-k13, line 143: I'm not sure if I'm doing something wrong here, the `.gcda` generated from old `runtest.py` is of the format:
```
./build/temp.linux-x86_64-3.10/build/src.linux-x86_64-3.10/numpy/core/src/umath/matmul.gcda
```

while the new one looks like:
```
./build/numpy/_core/_multiarray_umath.cpython-310-x86_64-linux-gnu.so.p/meson-generated_matmul.c.gcda
```

This means `lcov` is not working as expected. I'll take a look at this tomorrow, something off with the flags I'm guessing.

ganesh-k13: Meson supports coverage https://mesonbuild.com/howtox.html#producing-a-coverage-report, making life a lot easier.

ganesh-k13:
image

Steps to get this in NumPy:
1. `spin test --coverage --gcov`
2. `ninja coverage-html -C build` (This will be wrapped in a spin command/flag later)

spin/cmds/meson.py
```diff
@@ -109,8 +109,15 @@ def _meson_version_configured():
 @click.option(
     "-v", "--verbose", is_flag=True, help="Print detailed build and installation output"
 )
+@click.option(
+    "--gcov-build",
```
stefanv, line 113: For flags, I think recommended naming is `--gcov/--no-gcov`; see https://click.palletsprojects.com/en/8.1.x/options/#boolean-flags.

stefanv (commented):
Thanks @ganesh-k13! This looks good to me. I suggest we rename the flag to just `--gcov`, since it's already being applied to the build command:

```
spin build --gcov
```

It would be good to add a test for this. How about running the build with the `--gcov` flag, and then checking that the necessary report file exists afterwards?

spin/cmds/meson.py
rgommers: +1 for `--gcov` as the name. I don't see a point in adding `--no-gcov`, that doesn't do anything and is extra noise in `--help`. So I'd go for the second option documented in those Click docs under "If you really don’t want an off-switch".

stefanv: That works!

ganesh-k13: Added via [20276fc](https://github.com/scientific-python/spin/pull/146/commits/20276fc80612ca5a30c7de05c8c9eeac829c2eae)

jarrodmillman (approved):

.github/workflows/test.sh
```diff
@@ -13,7 +13,26 @@ prun cd example_pkg
 prun spin --version
 
 pip install meson-python ninja
-prun spin build
+
+
+# Test spin build + debug builds
+echo -e "${MAGENTA}Creating debug builds${NORMAL}"
```
stefanv, line 19: Please see the `ptest` command that has been introduced on `main`

.github/workflows/test.sh
```diff
@@ -13,7 +13,26 @@ prun cd example_pkg
...
+prun spin build --gcov
+echo -e "${MAGENTA}Did the build folder get generated?${NORMAL}"
+if [ ! -d "build" ] || [ ! -d "build-install" ]; then
+    echo -e "${RED}build and/or build-install folders did not get generated${NORMAL}"
```
stefanv, line 23: Makes me think we should also add `perror`!

spin/cmds/meson.py
```diff
@@ -215,8 +224,13 @@ def _get_configured_command(command_name):
     is_flag=True,
     help="Generate a coverage report of executed tests. An HTML copy of the report is written to `build/coverage`.",
 )
+@click.option(
+    "--gcov",
+    is_flag=True,
+    help="Enable C code coverage via gcov (requires GCC). gcov output goes to build/**/*.gc*",
```
stefanv, line 230: Should we advise, somewhere, on how to display those reports?

stefanv (commented):
Thanks, @ganesh-k13! This looks good overall. There were a few changes on `main` in the test suite that need to be accommodated for (sorry, I can also do it!), otherwise I think it's basically good to go.

.github/workflows/test.sh
ganesh-k13: Haha, I was actually thinking about the same thing but didn't want to introduce too many side changes. I'd be happy to refine it 👍

stefanv: 🙏

ganesh-k13: Added via [02e9e63](https://github.com/scientific-python/spin/pull/146/commits/02e9e6356ec55708210fa8702b410adc21a9ed02)

.github/workflows/test.sh
ganesh-k13: Refined in [02e9e63](https://github.com/scientific-python/spin/pull/146/commits/02e9e6356ec55708210fa8702b410adc21a9ed02)

spin/cmds/meson.py
ganesh-k13: Added via [62891ad](https://github.com/scientific-python/spin/pull/146/commits/62891ad92f394782c19070a13dab472d83c3274a)

stefanv (commented):
The PR looks great; thanks.

The only thing I can't figure out is what `ninja coverage*` refers to. Surely, that requires a ninja build file? Or is this generated by meson?

rgommers:
> The only thing I can't figure out is what `ninja coverage*` refers to. Surely, that requires a ninja build file? Or is this generated by meson?

Yes, `meson` uses `ninja`. The ninja file is located at `<build-dir>/build.ninja` (and `<build-dir>/.ninja_log` & co).

stefanv:
Right, I know that, but I don't think users by default do, and they certainly won't know that coverage commands are automatically generated. So it may be helpful to slightly expand the instructions to explain what is happening under the hood, and what to do (i.e. switch into build dir, then run command).

stefanv (approved):

stefanv:
Thanks, @ganesh-k13!

ganesh-k13:
Thanks all! When does the 0.9 release take place? No rush, just wanted to know so that I can rebase https://github.com/numpy/numpy/pull/24992 and the CI passes.

ganesh-k13:
I just realised `--generate-lcov-html` can also be part of `spin` itself instead of the individual downstreams implementing it as it's controlled by `ninja` and it's not a custom implementation of `lcov` as it was before in NumPy. Let me PR that when I'm back.

PR #150, opened 2024-01-11
Detect when a conflicting editable install exists

### LINKED ISSUE #143: Meson's editable installs overrides spin's own meson-based installation
I am not sure if this is necessarily a bug but it seems like something that should be discussed and people be made aware about. This concerns the development setup of scikit-image but may generalize to any project that uses spin.

meson-python supports [editable installs](https://meson-python.readthedocs.io/en/latest/how-to-guides/editable-installs.html). However, these seem to override spin's path manipulation completely. E.g. for scikit-image, first follow [Installing scikit-image for contributors (venv)](https://scikit-image.org/docs/dev/user_guide/install.html#venv) and then run

```sh
spin python -- -c 'import skimage; print(skimage)'
pip install -vv --no-build-isolation -e .
spin python -- -c 'import skimage; print(skimage)'
```
the first `spin python` will find the module in `REPO/build-install/...` as expected. However, after running `pip install -e` the module is different: `<module 'skimage' from 'REPO/skimage/__init__.py'>`. Running `pip uninstall -y scikit-image` will restore the original and expected behavior of spin.

So using both meson-python's editable installs and spin together has the consequence that most spin commands trigger two compilations: once from `spin build` and once from the editable hook during import.

Ideally spin could make sure that `build-install` is always used regardless of an editable hook being present. But not sure if that's easily possible or feasible...

See also this [thread on scikit-image's Zulip](https://skimage.zulipchat.com/#narrow/stream/181448-development/topic/dev.20installation/near/408931718) that triggered this issue.

stefanv:
Closes #143 

I'd like to check with @rgommers that this won't impact the workflow of those using editable packages. Do those users also use spin? If so, should we more explicitly accommodate editable installs, and simply skip the build step?

### DIFF
--- a/spin/cmds/meson.py
+++ b/spin/cmds/meson.py
@@ -36,6 +36,26 @@ def _set_pythonpath(quiet=False):
     site_packages = _get_site_packages()
     env = os.environ
 
+    cfg = get_config()
+    package = cfg.get("tool.spin.package", None)
+    if package:
+        import importlib_metadata
+
+        try:
+            dist = importlib_metadata.Distribution.from_name(package)
+            if dist.origin.dir_info.editable:
+                click.secho(
+                    f"Warning! An editable installation of `{package}` was detected.",
+                    fg="bright_red",
+                )
+                click.secho("Spin commands will pick up that version.", fg="bright_red")
+                click.secho(
+                    f"Try removing the other installation with `pip uninstall {package}`.",
+                    fg="bright_red",
+                )
+        except importlib_metadata.PackageNotFoundError:
+            pass
+
     if "PYTHONPATH" in env:
         env["PYTHONPATH"] = f"{site_packages}{os.pathsep}{env['PYTHONPATH']}"
     else:
--- a/spin/cmds/meson.py
+++ b/spin/cmds/meson.py
@@ -43,7 +43,7 @@ def _set_pythonpath(quiet=False):
 
         try:
             dist = importlib_metadata.Distribution.from_name(package)
-            if dist.origin.dir_info.editable:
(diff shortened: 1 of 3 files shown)


rgommers:
This seems right to me. It seems to me that use of `spin` will get too messy if it also tries to support editable builds. Not 100% sure, but it looks a little messy and it's not supported right now.

The NumPy docs spell it out very clearly right now (see note above https://numpy.org/devdocs/building/index.html#customizing-builds):

image

spin/cmds/meson.py
```diff
@@ -36,6 +36,26 @@ def _set_pythonpath(quiet=False):
...
     env = os.environ
 
+    cfg = get_config()
+    package = cfg.get("tool.spin.package", None)
+    if package:
+        import importlib_metadata
+
+        try:
+            dist = importlib_metadata.Distribution.from_name(package)
+            if dist.origin.dir_info.editable:
```
lagru, line 46: Should this also raise a different warning if a non-editable but installed distribution is detected? Possibly outside the directory expected by spin? That might lead to similar confusion.

Looks like an improvement otherwise! :+1:

stefanv: Fortunately, a non-editable installation does not cause problems.

The editable installs are particularly obnoxious, because their hooks are run before the PYTHONPATH takes effect.

stefanv:
> It seems to me that use of `spin` will get too messy if it also tries to support editable builds.

Since we now have a way to detect editable installs, we could potentially switch build and PYTHONPATH-setting to no-ops *if* the editable matches the current directory.

I'll look into that as a follow-up PR.

Issue #158, opened 2024-01-26
Generate C code coverage reports


ganesh-k13:
I just realised `--generate-lcov-html` can also be part of `spin` itself instead of the individual downstreams implementing it as it's controlled by `ninja` and it's not a custom implementation using `lcov` as it was before in NumPy. Let me PR that when I'm back.

_Originally posted by @ganesh-k13 in https://github.com/scientific-python/spin/issues/146#issuecomment-1907812452_

Issue #154, opened 2024-01-21
Improve unit testing using frameworks


ganesh-k13:
### Problem today
[test.sh](https://github.com/scientific-python/spin/blob/main/.github/workflows/test.sh) is growing today and will not be particularly clean when more commands are added.

### Proposal
Use any testing framework to make it easier to add more cases. I could think of two ways:
1. We stick with `bash` and use https://github.com/bats-core/bats-core
2. Migrate to pytest and use `subprocess` to call the spin commands

While the first method would probably be less code in the long term, it might not be as powerful as pytest. Parametrizing for instance would be easier.

I'd be happy to raise a PR to move to either or feel free to close this issue if we are ok to keep what we have for now.

### ToDo

- [ ] Add a pytest based testing framework
- [ ] Add tests for
  - [ ] `--version`
  - [ ] `build`
  - [ ] `test`
  - [ ] `PYTHONPATH` based tests
  - [ ] `docs`
  - [ ] `gdb`
  - [ ] Other misc tests in `test.sh`
- [ ] Depcrecate `test.sh`

stefanv:
The pytest route sounds good to me. I looked at the bash test frameworks before, and they're all somewhat sub-optimal.

We'll need to write some utility functions to check the outputs and error codes of commands, but that should be straightforward.

Thanks @ganesh-k13!

Shivaprabhath:
pytest might be a good choice. The ability to use parametrization can make it easier to add more test cases in the future. Bash has limited features compared to more pytest.

PR #159, opened 2024-01-26
Add ``--gcov-report`` flag to ``spin test``

ganesh-k13:
### Changes
- Added ``--gcov-report`` flag to ``spin test``

```sh
spin test --help                                                                                                    1 ↵ ganesh@ganesh-MS-7B86
Usage: spin test [OPTIONS] [PYTEST_ARGS]...
...
  --gcov-report [html|xml|text|sonarqube]
                                  Generate a coverage report for C Code and exit
...
```

#### Happy Path
```py
» spin test --gcov
...
» spin test --gcov-report html                                                                                    ganesh@ganesh-MS-7B86
Verifying installed packages for generating coverage reports...
$ ninja -C build -t targets all
Generating html coverage report...
$ ninja -C build coverage-html
Coverage report generated successfully and written to /home/ganesh/os/devpy/example_pkg/build/meson-logs/coveragereport/index.html
```
#### Failures
1. No build found
```py
» git clean -xdf                                                                                                           ganesh@ganesh-MS-7B86
Removing .spin/__pycache__/
Removing build-install/
Removing build/
» spin test --generate-gcov-report html                                                                                    ganesh@ganesh-MS-7B86
Verifying installed packages for generating coverage reports...
Error: `build` folder not found, cannot generate coverage reports. Generate coverage artefacts by running `spin test --gcov`
```
2. Gcovr, etc not installed
```py
» spin test --gcov-report html                                                                                   ganesh@ganesh-MS-7B86
Verifying installed packages for generating coverage reports...
$ ninja -C build -t targets all
Error: coverage-html is not supported... Ensure the following are installed: Gcovr/GenHTML, lcov and rerun `spin test --gcov`
```

#### ToDo

- [ ] UT

#### Misc
resolves https://github.com/scientific-python/spin/issues/158
continues https://github.com/scientific-python/spin/pull/146

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -3,5 +3,5 @@ import nox
 
 @nox.session
 def test(session: nox.Session) -> None:
-    session.install(".", "pytest", "build", "meson-python", "ninja")
+    session.install(".", "pytest", "build", "meson-python", "ninja", "gcovr")
     session.run("pytest", "spin", *session.posargs)
--- a/spin/cmds/meson.py
+++ b/spin/cmds/meson.py
@@ -1,11 +1,11 @@
 import contextlib
 import copy
-import enum
 import json
 import os
 import re
 import shutil
 import sys
+from enum import Enum
 from pathlib import Path
 
 import click
@@ -17,7 +17,7 @@ install_dir = "build-install"
 build_dir = "build"
 
 
-class GcovReports(str, enum.Enum):
+class GcovReportFormat(str, Enum):
     html = "html"
     xml = "xml"
     text = "text"
@@ -135,12 +135,28 @@ def _meson_version_configured():
         pass
 
 
-def _check_coverage_tool_installation(coverage_type: GcovReports):
+def _meson_coverage_configured() -> bool:
+    try:
+        build_options_fn = os.path.join(
+            "build", "meson-info", "intro-buildoptions.json"
+        )
+        with open(build_options_fn) as f:
+            build_options = json.load(f)
+        for b in build_options:
+            if (b["name"] == "b_coverage") and (b["value"] is True):
+                return True
+    except:
+        pass
+
+    return False
+
+
+def _check_coverage_tool_installation(coverage_type: GcovReportFormat):
     requirements = {  # https://github.com/mesonbuild/meson/blob/6e381714c7cb15877e2bcaa304b93c212252ada3/docs/markdown/Unit-tests.md?plain=1#L49-L62
-        GcovReports.html: ["Gcovr/GenHTML", "lcov"],
-        GcovReports.xml: ["Gcovr (version 3.3 or higher)"],
-        GcovReports.text: ["Gcovr (version 3.3 or higher)"],
-        GcovReports.sonarqube: ["Gcovr (version 4.2 or higher)"],
+        GcovReportFormat.html: ["Gcovr/GenHTML", "lcov"],
+        GcovReportFormat.xml: ["Gcovr (version 3.3 or higher)"],
+        GcovReportFormat.text: ["Gcovr (version 3.3 or higher)"],
+        GcovReportFormat.sonarqube: ["Gcovr (version 4.2 or higher)"],
     }
 
     # First check the presence of a valid build
@@ -153,8 +169,8 @@ def _check_coverage_tool_installation(coverage_type: GcovReports):
     debug_files = Path(build_dir).rglob("*.gcno")
     if len(list(debug_files)) == 0:
         raise click.ClickException(
-            "debug build not found, cannot generate coverage reports. "
-            "Generate coverage artefacts by running `spin test --gcov`"
+            "Debug build not found, cannot generate coverage reports.\n\n"
+            "Please rebuild using `spin build --clean --gcov` first."
         )
 
     # Verify the tools are installed prior to the build
@@ -176,18 +192,7 @@ def _check_coverage_tool_installation(coverage_type: GcovReports):
 @click.option(
     "--gcov",
     is_flag=True,
-    help="""Enable C code coverage via `gcov`.
-
-    The meson-generated `build/build.ninja` has targets for compiling
-    coverage reports.
-
-    E.g., to build an HTML report, in the `build` directory run
-     `ninja coverage-html`.
-
-    To see a list all supported formats, run
-    `ninja -t targets | grep coverage-`.
-
-    Also see https://mesonbuild.com/howtox.html#producing-a-coverage-report.""",
+    help="Enable C code coverage using `gcov`. Use `spin test --gcov` to generate reports.",
 )
 @click.argument("meson_args", nargs=-1)
 def build(meson_args, jobs=None, clean=False, verbose=False, gcov=False, quiet=False):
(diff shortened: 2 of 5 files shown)


.github/workflows/test.sh
```diff
@@ -89,9 +89,42 @@ if [[ $PLATFORM == linux || $PLATFORM == darwin ]]; then
...
+prun spin test --gcov
+matching_files=$(find . -type f -name "*.gcda")
+if [ -z "$matching_files" ]; then
+    perror Coverage files did not get generated
+    exit 1
+else
+    echo "Yes"
+fi
+
+# TODO: Test other coverage reports
```
ganesh-k13, line 106: TODO: I'll cover this as a part of https://github.com/scientific-python/spin/issues/154 to avoid bloat here

spin/cmds/meson.py
```diff
@@ -124,6 +134,31 @@ def _meson_version_configured():
         pass
 
 
+def _check_coverage_tool_installation(coverage_type: GcovReports):
+    requirements = {  # https://github.com/mesonbuild/meson/blob/6e381714c7cb15877e2bcaa304b93c212252ada3/docs/markdown/Unit-tests.md?plain=1#L49-L62
+        GcovReports.html: ["Gcovr/GenHTML", "lcov"],
```
ganesh-k13, line 139: Good to know this works :) https://github.com/scientific-python/spin/actions/runs/7677258874/job/20925774260?pr=159#step:5:212

stefanv:
Thanks, Ganesh, I've done a brief pass and all looks good. There are a few small tweaks I'd like to make, which I'll do tomorrow morning.

Instead of generate-gcov-report, perhaps the shorter gcov-report? Happy to make that tweak myself, just wanted to make sure you don't object.

ganesh-k13:
Thanks for the review @stefanv , I have renamed the flag to `--gcov-report` and updated the description.

spin/cmds/meson.py
```diff
@@ -124,6 +134,31 @@ def _meson_version_configured():
...
+            "`build` folder not found, cannot generate coverage reports. "
+            "Generate coverage artefacts by running `spin test --gcov`"
+        )
+
+    # Verify the tools are installed prior to the build
+    p = _run(["ninja", "-C", build_dir, "-t", "targets", "all"], output=False)
+    if f"coverage-{coverage_type.value}" not in p.stdout.decode("ascii"):
+        raise click.ClickException(
+            f"coverage-{coverage_type.value} is not supported... "
+            f"Ensure the following are installed: {', '.join(requirements[coverage_type])} "
```
stefanv, line 157: This error is raised when `spin build; spin test --gcov-report`. Needs to ensure that build was run with `--gcov`.

stefanv:
After looking at this some more, I'm wondering if the `gcov-report` option is associated with the right command. It seems unrelated to the test options, and rely on the build having been performed with `--gcov`.

Also, doing `spin build`, and then `spin build --gcov` isn't enough. You need `spin build --clean --gcov`.

So, I'd suggest moving `--gcov-report` to build, setting `html` as the default, automatically enabling `--gcov` when `--gcov-report` is set, and also detecting when the existing build wasn't run with `--gcov` (you can check the `build/build.ninja` file for `coverage-*` commands).

rgommers:
> Also, doing `spin build`, and then `spin build --gcov` isn't enough. You need `spin build --clean --gcov`.

I'll note that this isn't specific to `--gcov`, it applies to anything that translates to a compile/link flag. `--werror` and `--asan`, which SciPy's `dev.py` has any would also be useful for `spin`, are good examples.



> So, I'd suggest moving `--gcov-report` to build, setting `html` as the default, automatically enabling `--gcov` when `--gcov-report` is set

I'm not sure I understand that suggestion, or maybe I'm missing your point. You typically build from scratch once, and run tests many times when you're developing. You don't want a coverage report every time you run tests, so it shouldn't be done automatically. But conceptually, you want the coverage report for the tests you've just run I think, through a "give me the gcov coverage report". So it could be integrated with `test`, or possibly standalone - but it doesn't belong with `build`.

stefanv:
That was my original understanding, but then I couldn't see where the gcov outputs are generated by pytest. I clearly missed it, so I'll take another look.

rgommers:
> That was my original understanding, but then I couldn't see where the gcov outputs are generated by pytest. I clearly missed it, so I'll take another look.

It's a separate command, unrelated to `pytest`. That command does need to be run after the test run of interest. E.g., see https://github.com/scipy/scipy/blob/maintenance/1.10.x/runtests.py#L510-L533.

spin/cmds/meson.py
```diff
@@ -261,11 +300,20 @@ def _get_configured_command(command_name):
     "--gcov",
     is_flag=True,
     help="Enable C code coverage via `gcov`. `gcov` output goes to `build/**/*.gc*`. "
-    "Reports can be generated using `ninja coverage*` commands. "
+    "Reports can be generated using `spin test --generate-gcov-report *` command. "
```
jarrodmillman, line 303:
```suggestion
    "Reports can be generated using `spin test --gcov-report *` command. "
```

spin/cmds/meson.py
ganesh-k13: Yep, shall I add a dependency to call `spin build --gcov` in case it's missing?

spin/tests/test_build_cmds.py
```diff
@@ -58,6 +58,34 @@ def test_debug_builds():
...
+
+    coverage_files = Path(".").rglob("*.gcda")
+    assert len(list(coverage_files)) != 0, "coverage files not generated for gcov build"
+
+
+@pytest.mark.parametrize(
+    "report_type,output_file",
+    [
+        ("html", Path("coveragereport/index.html")),
+        ("xml", Path("coverage.xml")),
```
ganesh-k13, line 73: Ah `gcovr` is missing, hence the tests are failing. Need to add to nox env.

noxfile.py
```diff
@@ -3,5 +3,5 @@
 
 @nox.session
 def test(session: nox.Session) -> None:
-    session.install(".", "pytest", "build", "meson-python", "ninja")
+    session.install(".", "pytest", "build", "meson-python", "ninja", "gcovr")
```
ganesh-k13, line 6: @stefanv / @jarrodmillman , should this be part of a `test_requirements.yaml` for depandabot to pick up and auto update or this is ok?

spin/cmds/meson.py
ganesh-k13: Ah ok, I get the problem. We need to check for coverage artifacts. Let me add it.

ganesh-k13: Added checks and UT in f2c21eb9d658e5a22dbd991bd217dd52030f0d12

stefanv:
@ganesh-k13 Sorry for the delay working on this. I've reworked the logic a bit, as detailed in the last commit message. Take a look, and see if you approve of what I've done?

jarrodmillman (approved):

ganesh-k13 (commented):
Thanks, looks good! I cannot approve as I'm the author, but looks good to go. I'll port this to NumPy and hopefully enable code cov reports in the CI soon.

PR #170, opened 2024-03-15
Document how to run test inside of LLDB

stefanv:


seberg:
Yeah, I don't have a great thought and think this is fine, thanks!  Having a short-hand without any options seems not a lot more useful.

stefanv:
FWIW, it's also possible to add a custom command to NumPy that encodes the semantics you'd like to see. Since it's not meant to be general, the design would be quite flexible.

PR #190, opened 2024-05-14
Remove editable install caveat from README

lesteve:
I believe this has been fixed in https://github.com/scientific-python/spin/pull/155, released in 0.9, or maybe I am missing something?


stefanv:
Thanks, that's right!

PR #192, opened 2024-05-14
Improve debug printing for Meson editable installs

stefanv:


spin/cmds/pip.py
```diff
@@ -36,6 +36,11 @@ def install(pip_args, verbose, editable):
     pip_args += ["--no-build-isolation"]
     if editable:
         pip_args += ["--editable"]
+        if verbose:
```
lesteve, line 39: Thanks for this PR!

I find it a tiny bit weird to mix two slightly different things:
- verbose for pip commands, I want to see the output of the `pip` command
- verbose for meson editable install: I want to see the compilation steps when importing the package

Maybe I want to see the compilation output the first time (pip verbose) but not on import (meson editable verbose)

A separate parameter like `--meson-editable-verbose` would maybe makes sense?

stefanv: Sure, let's split the options.

stefanv: OK, done.

stefanv:
> I will trust you on the naming, I proposed `--meson-editable-verbose` initially, but I would agree that `--verbose-import` is more easily understandable for most people and also shorter.

You know how it is with naming. There's no right or wrong, but everyone's got an opinion :) I chose this to keep it short, and in case we could retrofit it in the future to CMake and other build types.

lesteve:
Thanks for this! Just curious, what's the rough timeline for the 0.10 release? This is mostly so that I remember to update https://github.com/scikit-learn/scikit-learn/pull/29012.

stefanv:
0.10 should be out by tomorrow.

lesteve:
Great thanks!

stefanv:
Release is out.

lesteve:
Thanks!

PR #194, opened 2024-05-14
`dir_info` metadata field does not exist for conda installs

### LINKED ISSUE #193: spin 0.9 update causes AttributeError in numpy CI
It looks like this broke one of the numpy CI jobs: https://github.com/numpy/numpy/actions/runs/9087080853/job/24974107681?pr=26436

```
Error: 'types.SimpleNamespace' object has no attribute 'dir_info'; aborting.
```

And indeed there's a new access of a `dir_info` attribute added in #155.

Clues about how to fix this in the numpy CI would be appreciated if this isn't a spin bug.

stefanv:
Closes #193

### DIFF
--- a/spin/cmds/meson.py
+++ b/spin/cmds/meson.py
@@ -45,7 +45,9 @@ def _editable_install_path(distname):
     except importlib_metadata.PackageNotFoundError:
         return None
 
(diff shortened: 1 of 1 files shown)


stefanv:
@jarrodmillman This change warrants a point release.

Issue #209, opened 2024-06-19
META: Desired test behavior


stefanv:
The current spin release's test command does not work for the majority of packages in the ecosystem. This is related to support added for editable installs, and the subsequent decision to [add --pyargs to each pytest invocation](https://github.com/scientific-python/spin/pull/191).

This led to related problems, including #208, #204, #203, and https://github.com/numpy/numpy/issues/26525

This issue is an attempt to write up the requirements for `spin test` for various configurations, which is a cross-product of:

- **Installation methodology**: Editable install vs built into `build-install`.
- **Source or classic layout**: Are sources in `src/` or `packagename/`?
- **Tests bundled / unbundled**: Are tests only in `./tests`, `./packagename/**/tests`, or included as part of the package?
- **conftest.py**: Does the package include a `conftest.py` (which may, perhaps, import the package)?

I was hoping to find a Single Invocation of `pytest` that would handle all of these cases, but I suspect that that may not be possible. Instead, I am now creating test packages of the various flavors to see to what degree we'd have to customize the pytest call.

In a future comment, I may document the implications of each of the various configurations.

The biggest obstacle I have right now is how to support filename test specications, such as

```
spin test numpy/linalg/tests/test_linalg.py
spin test tests/some/test_module.py
```

for the case where tests are not shipped along with the package.

Issue #208, opened 2024-06-10
spin test with arguments does not work when `spin test` works alone


adam2392:
I am running into some weird behavior when upgrading spin recently. I am running spin on GH actions and locally: https://github.com/neurodata/scikit-tree/actions/runs/9441764126/job/26003160931

```
(sktree) (base) adam2392@arm64-apple-darwin20 scikit-tree % ./spin test -t ./sktree --coverage --gcov -- -m "slowtest"
Invoking `build` prior to running tests:
$ meson setup build --prefix=/usr -Db_coverage=true --reconfigure
The Meson build system
Version: 1.4.0
Source dir: /Users/adam2392/Documents/scikit-tree
Build dir: /Users/adam2392/Documents/scikit-tree/build
Build type: native build
Project name: scikit-tree
Project version: 0.8.0dev0
C compiler for the host machine: arm64-apple-darwin20.0.0-clang (clang 15.0.7 "clang version 15.0.7")
C linker for the host machine: arm64-apple-darwin20.0.0-clang ld64 609
C++ compiler for the host machine: arm64-apple-darwin20.0.0-clang++ (clang 15.0.7 "clang version 15.0.7")
C++ linker for the host machine: arm64-apple-darwin20.0.0-clang++ ld64 609
Cython compiler for the host machine: cython (cython 3.0.10)
Host machine cpu family: aarch64
Host machine cpu: aarch64
Compiler for C supports arguments -Wno-unused-but-set-variable: YES (cached)
Compiler for C supports arguments -Wno-unused-function: YES (cached)
Compiler for C supports arguments -Wno-conversion: YES (cached)
Compiler for C supports arguments -Wno-misleading-indentation: YES (cached)
Compiler for C supports arguments -DNPY_NO_DEPRECATED_API=NPY_1_7_API_VERSION: YES (cached)
Library m found: YES
Program cython found: YES (/Users/adam2392/miniforge3/envs/sktree/bin/cython)
Program python3 found: YES (/Users/adam2392/miniforge3/envs/sktree/bin/python3.9)
Message: /Users/adam2392/miniforge3/envs/sktree/bin/python3.9
Message: /usr/lib/python3.9/site-packages/
Dependency OpenMP found: YES 5.0 (cached)
Build targets in project: 19

scikit-tree 0.8.0dev0

  User defined options
    prefix    : /usr
    b_coverage: true

[...]

Issue #204, opened 2024-05-29
`--pyargs` not used if `-k` is used


oscarbenjamin:
I'm seeing this problem with a src-layout in both an editable install and also a plain `spin test`. If I pass `-k` through to `pytest` then `--pyargs` is not also passed:
```console
$ spin test -- -k test_mfpz_mod_poly
...
$ /home/oscar/.pyenv/versions/3.11.3/envs/python-flint-3.11/bin/python3.11 -P -m pytest -k test_mfpz_mod_poly
======================================== test session starts ========================================
platform linux -- Python 3.11.3, pytest-7.4.3, pluggy-1.2.0
rootdir: /home/oscar/current/active/python-flint
plugins: hypothesis-6.84.2, cov-4.1.0, split-0.8.1, timeout-2.1.0, doctestplus-1.0.0, xdist-3.3.1
collected 49 items / 1 error / 49 deselected / 0 selected                                           

============================================== ERRORS ===============================================
____________________________ ERROR collecting src/flint/test/test_all.py ____________________________
import file mismatch:
imported module 'flint.test.test_all' has this __file__ attribute:
  /home/oscar/current/active/python-flint/build-install/usr/lib/python3.11/site-packages/flint/test/test_all.py
which is not the same as the test file we want to collect:
  /home/oscar/current/active/python-flint/src/flint/test/test_all.py
HINT: remove __pycache__ / .pyc files and/or use a unique basename for your test file modules
====================================== short test summary info ======================================
ERROR src/flint/test/test_all.py
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
================================== 49 deselected, 1 error in 0.35s ==================================
```
A plain `spin test` works because it uses `--pyargs`:
```console
$ spin test
...
$ /home/oscar/.pyenv/versions/3.11.3/envs/python-flint-3.11/bin/python3.11 -P -m pytest --pyargs flint
======================================== test session starts ========================================
platform linux -- Python 3.11.3, pytest-7.4.3, pluggy-1.2.0
rootdir: /home/oscar/current/active/python-flint
plugins: hypothesis-6.84.2, cov-4.1.0, split-0.8.1, timeout-2.1.0, doctestplus-1.0.0, xdist-3.3.1
collected 49 items                                                                                  

[...]

Issue #203, opened 2024-05-28
test `--pyargs` invocation breaks NumPy CI on Windows


stefanv:
Resulting from #191

stefanv:
Closed by #213

PR #211, opened 2024-06-21
Fix testing for editable and non-editable installs alike

stefanv:
Work to address #209, #203, #204, #208

### DIFF
--- a/example_pkg/example_pkg/meson.build
+++ b/example_pkg/example_pkg/meson.build
@@ -6,7 +6,8 @@ py.extension_module(
 )
 
 python_sources = [
-  '__init__.py'
+  '__init__.py',
+  'conftest.py'
 ]
 
 py.install_sources(
@@ -14,4 +15,5 @@ py.install_sources(
   subdir: 'example_pkg'
 )
 
-install_subdir('tests', install_dir: py.get_install_dir() / 'example_pkg/tests')
+install_subdir('submodule', install_dir: py.get_install_dir() / 'example_pkg')
+install_subdir('tests', install_dir: py.get_install_dir() / 'example_pkg')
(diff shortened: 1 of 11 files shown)


stefanv:
@ngoldbaum This should fix the problems with NumPy, hopefully. There's still one problem due to the way NumPy overrides this command, such that `spin test numpy.linalg` does not work (vs `spin test -t numpy.linalg`, which does). It'd be an easy fix on the NumPy side, though.

stefanv:
@jarrodmillman Given that the current test command is broken, I recommend we do this as a point release, but we could also just do a 0.11.

ngoldbaum:
> It'd be an easy fix on the NumPy side, though.

Sounds reasonable. Care to elaborate a little bit?

stefanv:
Sure! The user is supposed to supply tests as `spin test -t test.name`, but as a convenience we also allow `spin test.name`. To detect this case, we check for a single argument (maybe there's a better way, but that's what happens right now), and NumPy modifies the arguments of `spin test foo.bar` to look something like `-m 'not slow' foo.bar`. The simplest fix is to take a single argument test specification and pass it as the `tests` argument, instead of as part of `pytest_args`.

ngoldbaum:
I tried testing this after implementing the suggested change to numpy's spin configuration, but I'm still getting an error trying to do `spin build` on this branch of `spin`:

```
± spin test numpy.linalg
Error: 'NoneType' object has no attribute 'dir_info'; aborting.
```

This is coming from here: https://github.com/scientific-python/spin/blob/b705a480febbe145d461e5a36d9456c9bb13f442/spin/cmds/meson.py#L49

I don't know enough about how `importlib_metadata` works to know why `Distribution.from_name('numpy').origin` is `None` on my system and python install:

```
In [1]: import importlib_metadata

In [2]: print(importlib_metadata.Distribution.from_name('numpy').origin)
None
```

stefanv:
OK, I thought we patched that; let me check why that error is still present for you.

stefanv:
https://github.com/scientific-python/spin/pull/212

ngoldbaum:
See https://github.com/numpy/numpy/pull/26813

stefanv:
That wasn't supposed to close. Reopened as #213

PR #212, opened 2024-06-28
Do not fail editable install detection when only system install is present

stefanv:
Closes #211


ngoldbaum:
This fixes the error I saw in the other issue, thanks!

PR #221, opened 2024-07-13
Allow custom build directory

### LINKED ISSUE #171: Facilitate parallel build directories for meson-enabled commands?
While making scikit-image compatible with NumPy 2 I often had to switch back and forth between 2 environments with NumPy 1.x and 2 installed. I had to use `spin build --clean` each time I switched which even with ccache available took a bit of time. From the initial introduction of meson and it "breaking" inplace editable installs I remember that one of the arguments was that meson would allow having multiple build directories in parallel. Perhaps I missed an obvious way to achieve this with existing capabilities of spin?

If not and going forward, can we facilitate this workflow a bit more in spin? For my scenario I imagined something like a `-t, --target TARGET` switch. If not specified the default setup with `build/skimage` and `build-install` is used. With `-t np2` meson-enabled spin commands would use `build/np2/skimage` and maybe `build-install/np2`? 

UI and paths are of course up for discussion but I think it get's the workflow across that I'd like to have. It would enable faster switching and debugging with multiple environments that require recompilation.

stefanv:
Closes #171

- [X] Add `-C` flag to meson commands
- [x] Add tests and verify that compilation happens in right place

### DIFF
--- a/spin/cmds/meson.py
+++ b/spin/cmds/meson.py
@@ -234,7 +234,9 @@ build_dir_option = click.option(
     "-C",
     "--build-dir",
     default="build",
-    help="Meson build directory; package is installed into './{build-dir}-install'",
+    show_envvar=True,
+    envvar="SPIN_BUILD_DIR",
+    help="Meson build directory; package is installed into './{build-dir}-install'.",
 )
 
 
@@ -279,13 +281,14 @@ def build(
 
       CFLAGS="-O0 -g" spin build
 
-    Build into a different build/build-install directory:
+    Build into a different build/build-install directory via the
+    `-C/--build-dir` flag:
 
       spin build -C build-for-feature-x
 
     This feature is useful in combination with a shell alias, e.g.:
 
-      $ alias spin-clang="spin -C build-clang"
+      $ alias spin-clang="SPIN_BUILD_DIR=build-clang CC=clang spin"
 
     Which can then be used to build (`spin-clang build`), to test (`spin-clang test ...`), etc.
 
@@ -350,7 +353,9 @@ def build(
             "-C",
             build_dir,
             "--destdir",
-            f"../{install_dir}",
+            install_dir
+            if os.path.isabs(install_dir)
+            else os.path.join("..", install_dir),
         ],
         output=(not quiet) and verbose,
     )
--- a/spin/cmds/meson.py
+++ b/spin/cmds/meson.py
@@ -263,7 +263,8 @@ def build(
 ):
     """🔧 Build package with Meson/ninja
 
-    The package is installed to `build-install`.
+    The package is installed to `build-install` (unless a different
+    build directory is specified with `-C`).
 
     MESON_ARGS are passed through e.g.:
 
@@ -277,6 +278,17 @@ def build(
     or set CFLAGS appropriately:
 
       CFLAGS="-O0 -g" spin build
+
+    Build into a different build/build-install directory:
+
+      spin build -C build-for-feature-x
+
+    This feature is useful in combination with a shell alias, e.g.:
+
+      $ alias spin-clang="spin -C build-clang"
+
+    Which can then be used to build (`spin-clang build`), to test (`spin-clang test ...`), etc.
+
     """
     install_dir = _get_install_dir(build_dir)
     cfg = get_config()
--- a/spin/cmds/meson.py
+++ b/spin/cmds/meson.py
@@ -172,9 +172,9 @@ def _meson_version():
         pass
 
 
-def _meson_version_configured():
+def _meson_version_configured(build_dir):
     try:
-        meson_info_fn = os.path.join("build", "meson-info", "meson-info.json")
+        meson_info_fn = os.path.join(build_dir, "meson-info", "meson-info.json")
         with open(meson_info_fn) as f:
             meson_info = json.load(f)
         return meson_info["meson_version"]["full"]
@@ -316,7 +316,7 @@ def build(
         if os.path.isdir(install_dir):
             shutil.rmtree(install_dir)
 
-    if not (os.path.exists(build_dir) and _meson_version_configured()):
+    if not (os.path.exists(build_dir) and _meson_version_configured(build_dir)):
         p = _run(setup_cmd, sys_exit=False, output=not quiet)
         if p.returncode != 0:
             raise RuntimeError(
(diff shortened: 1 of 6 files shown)


stefanv:
@lagru Ready for testing.

spin/cmds/meson.py
```diff
@@ -257,7 +278,19 @@ def build(
...
+
+    Build into a different build/build-install directory:
+
+      spin build -C build-for-feature-x
+
+    This feature is useful in combination with a shell alias, e.g.:
+
+      $ alias spin-clang="spin -C build-clang"
+
+    Which can then be used to build (`spin-clang build`), to test (`spin-clang test ...`), etc.
```
lagru, line 290: The name `build-clang` might be a bit misleading. To me it implies that this feature can somehow be used to choose a different compiler (clang). As I understand it, you'd need to update the underlying "build" command / configuration which would then apply regardless of build directory? Or am I missing something?

spin/cmds/meson.py
```diff
@@ -257,7 +278,19 @@ def build(
...
 
       CFLAGS="-O0 -g" spin build
+
+    Build into a different build/build-install directory:
+
+      spin build -C build-for-feature-x
+
+    This feature is useful in combination with a shell alias, e.g.:
+
+      $ alias spin-clang="spin -C build-clang"
```
lagru, line 288: `spin -C ${SOME_DIR} command` seems to fail. I think this is because the option is added ad subcommand level. So this alias won't work currently. I think the option would need to be added to the top-level `spin` command and then passed down via the [context](https://click.palletsprojects.com/en/8.1.x/complex/#contexts).

spin/cmds/meson.py
```diff
@@ -226,6 +230,14 @@ def _check_coverage_tool_installation(coverage_type: GcovReportFormat):
         )
 
 
+build_dir_option = click.option(
+    "-C",
+    "--build-dir",
+    default="build",
+    help="Meson build directory; package is installed into './{build-dir}-install'",
```
lagru, line 237: This help text implies that subdirectories are not supported. And indeed, running
```sh
spin build -C build/testspin
spin python -C build/testspin
```
fails with
```plain
FileNotFoundError: No `site-packages` or `dist-packages` found under `build/testspin-install`
```

I'd definitely appreciate support for using a full path (sub directories, absolute path) here. That would allow targeting a generic `build/...` which is way easier to cover with `.gitignore` than `{build-dir}` which can have any name. It would also allow such cool things as
```
spin build -c /tmp/checkout-pr-feature
```

As it's currently implemented, this will quickly clog up valuable space in the top-level dir.

---

Another thought is to always use `build-intall` regardless of the `-C´ option and make meson update that directory with the cached build in `${build-dir}` when needed.

lagru (commented):
Thanks @stefanv! I played around with it locally and came up with a few limitations. Would be good to iterate further on the API and behavior. :)

spin/cmds/meson.py
stefanv: Right, may have to rethink that. We don't have global context in spin.

stefanv:
@lagru Please take another look; the build directory can now be absolute, and can be set using `SPIN_BUILD_DIR`, which is helpful for creating an alias. I also updated the alias to be a better example, showing how to set the compiler to clang and build into a different build directory using clang.

spin/cmds/meson.py
lagru: Hmm, with the most recent version
```
spin build -C build/testspin
spin python -C build/testspin
```
still fails with the same error:
```plain
Traceback (most recent call last):
  File "/home/lg/.local/lib/micromamba/envs/skidev312/lib/python3.12/site-packages/spin/cmds/meson.py", line 738, in python
    p = _set_pythonpath(build_dir)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lg/.local/lib/micromamba/envs/skidev312/lib/python3.12/site-packages/spin/cmds/meson.py", line 100, in _set_pythonpath
    site_packages = _get_site_packages(build_dir)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lg/.local/lib/micromamba/envs/skidev312/lib/python3.12/site-packages/spin/cmds/meson.py", line 160, in _get_site_packages
    raise FileNotFoundError(
FileNotFoundError: No `site-packages` or `dist-packages` found under `build/testspin-install`
```

It works for an absolute path such as `/tmp/testspin` though! Did you test the relative path successfully on your side? Then it might be something with my setup..

lagru: Is the leading `..` in `--destdir ../build/testspin-install` the culprit perhaps? (I updated the above traceback to include the full output)

stefanv: Ah, you want nested relative dirs; I'll fix that use case too.

lagru (approved):
Thanks. Absolute, relative and nested-relative paths seem to work perfectly now. I'm not familiar with the code base so take my approval with a grain of salt. ;)

jarrodmillman (approved):

oscarbenjamin:
Nice!

I was just about to suggest this feature. I just gave it a quick er... spin and it seems to work exactly as I wanted.

One suggestion though:

When you run `meson setup build-dir` it adds a `.gitignore` to the build directory. I already have a `.gitignore` entry for the `build-install` directory but if I do:
```bash
meson setup build-release -Dbuildtype=release
spin -C build-release run ...
```
Then the `build-release-install` directory is created but it does not have any `.gitignore` file. Perhaps `spin` could create the directory and add the `.gitignore` file before running `meson install`.

stefanv:
@oscarbenjamin What you request sounds very reasonable, but it's already the way my setup behaves:

```sh
$ spin build -C build-xyz && ls -al build-xyz/.gitignore
-rw-r--r--. 1 stefan stefan 92 Sep  2 21:05 build-xyz/.gitignore
```

I see the same behavior when adding the `-Dbuildtype=release` flag.

oscarbenjamin:
> it's already the way my setup behaves:

I've just checked again and confirm that the `.gitignore` is not added here:
```console
$ spin build -C build-foo -- --pkg-config-path=.local/lib/pkgconfig -Dadd_flint_rpath=true
...
$ git status
On branch pr_nmod_ctx
Your branch is up-to-date with 'origin/pr_nmod_ctx'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	build-abc-install/
	build-foo-install/
	build-xyz-install/

nothing added to commit but untracked files present (use "git add" to track)
```
Is it meson that adds it in the `meson install` step?

I didn't expect that `meson install` would do that so it didn't seem surprising that the `.gitignore` was not being added.

Issue #226, opened 2024-07-29
Add a global `--dry-run` to allow previewing a command


betatim:
What do people think of having a "dry run" mode for spin commands? Something that would print out all the commands that would be executed as well as other things the command would do.

I like having a "dry run" mode for tools like spin as a solution to the problem that when I encounter a "random project from the internet" that doesn't provide me explicit commands/steps I recognise to install it/setup for development and instead says "just run `./foo bar --baz`" I get worried about what exactly it will do. This is less of a problem for projects I know or use frequently, more for ones I've not used before or don't know. Which is exactly when the convenience of a tool like spin is the most useful.

My paranoia comes from having had projects `brew install` things, do other global reconfigurations, install packages into my current environment, etc, etc. Partly they are hard to notice in the moment, partly they are hard to undo and partly it is just annoying to have that happen.

Hence the desire for a "dry run" where I can see what something will do before I run it.

I Had a quickly look around and it looks like `spin.cmds.util.run` could be a central place to hook into and make it print the command instead of actually executing it. But there are also commands like https://github.com/scientific-python/spin/blob/2216c272881b9e37e4b414804bd347a555b5b0a5/spin/cmds/meson.py#L281 which call `rmtree` directly. So you'd have to add support for "dry run" to each command. This means it is a bit more tricky to do/each command would have to add support itself.

WDYT?

stefanv:
I like the idea, but I suspect you're going to run into issues with commands that follow the "Do A" - "Check that A was done" - "Do B" flow.

Two other options we can consider:

- Add the ability for a command to print out its own Python code, in a form that can be separately executed
- Ask for confirmation before each command executed (but this doesn't help for the case in which Python code modifies things)

betatim:
A third thing to do could be to provide some helper stuffs so command authors can easily write a command that prints a message "This command will create a new conda env called 'dev', install packages into it (big download), and then run the 'foo' executable." and then asks "Do it? [Y/n]". Together with a `-y` flag to skip the message.

At least it looks like doing anything global will only catch some of the cases and really you need to appeal to command authors to write nice commands.

stefanv:
That last bit (query with y flag) would be very easy to add.

stefanv:
See https://click.palletsprojects.com/en/8.1.x/prompts/#confirmation-prompts

stefanv:
@betatim See also this PoC for introspecting command code: https://github.com/scientific-python/spin/pull/228

[...]

PR #228, opened 2024-08-15
Add introspection command

stefanv:
Partially addresses #226

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,15 +86,15 @@ python -m spin
 Available as `spin.cmds.meson.*`.
 
 ```
-  build    🔧 Build package with Meson/ninja
-  ipython  💻 Launch IPython shell with PYTHONPATH set
-  python   🐍 Launch Python shell with PYTHONPATH set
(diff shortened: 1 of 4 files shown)


betatim:
Looks interesting. Certainly worth exploring and using a bit. For short commands it'll be great to see the source, for ones with a long implementation the "it is defined here xxx" is a good idea! Can we make it the actual file path instead? I'd have to think for a second or two to figure out how to open the file corresponding to the module path (or what ever it is called).

Another random thought: should the line "the blah command is defined in foo" appear at the end of the print out? Was thinking that if the command's code is too long for me to want to view it in the terminal "as is", then probably the top has scrolled off the screen as well. So putting the pointer at the end would make it easier to discover. Or maybe if the code is more than X lines (bike shed incoming :D) just display "This command is quite long, read its full source by opening xxxx in an editor"?

stefanv:
@betatim The file location is now printed, and it appears *after* the code.

PR #238, opened 2024-10-07
Update `spin` to make it a drop in replacement of `scipy/dev.py`

czgdp1807:
Regarding context - We are shifting to `spin` for building, testing SciPy. This PR is for updating `spin` to make it work with the SciPy building process. One explanation for the update is given below. That is significant. Rest for the smaller updates in `spin` I will provide explanation in the code review when I will complete the work.

In SciPy each step i.e., `compile`, `setup` and `install` needs arguments from `meson_args`. You can't pass everything from `meson_args` tuple directly into the first command. So I added 3 keys, `"compile"`, `"setup"` and `"install"` to differentiate which meson argument from SciPy to pass in which step.

Relevant PR from SciPy - https://github.com/scipy/scipy/pull/21674
~Relevant PR from NumPy - https://github.com/numpy/numpy/pull/27606~

### DIFF
--- a/spin/cmds/meson.py
+++ b/spin/cmds/meson.py
@@ -319,12 +319,12 @@ def build(
             )
         return
 
-    meson_args = list(meson_args)
+    meson_args_setup = list(meson_args["setup"])
 
     if gcov:
-        meson_args = meson_args + ["-Db_coverage=true"]
+        meson_args_setup = meson_args_setup + ["-Db_coverage=true"]
 
-    setup_cmd = _meson_cli() + ["setup", build_dir, f"--prefix={prefix}"] + meson_args
+    setup_cmd = _meson_cli() + ["setup", build_dir, "--prefix=/usr"] + meson_args_setup
 
     if clean:
         print(f"Removing `{build_dir}`")
@@ -351,15 +351,19 @@ def build(
 
         # Any other conditions that warrant a reconfigure?
 
+    meson_args_compile = list(meson_args["compile"])
     compile_flags = ["-v"] if verbose else []
-    if jobs:
-        compile_flags += ["-j", str(jobs)]
+    if "jobs" in meson_args:
+        jobs = meson_args["jobs"]
+    compile_flags += ["-j", str(jobs)]
 
     p = _run(
-        _meson_cli() + ["compile"] + compile_flags + ["-C", build_dir],
+        _meson_cli() + ["compile"] + compile_flags + ["-C", build_dir] ,
         sys_exit=True,
         output=not quiet,
     )
+
+    meson_args_install = list(meson_args["install"])
     p = _run(
         _meson_cli()
         + [
@@ -371,7 +375,7 @@ def build(
             install_dir
             if os.path.isabs(install_dir)
             else os.path.relpath(abs_install_dir, abs_build_dir),
-        ],
+        ] + meson_args_install,
         output=(not quiet) and verbose,
     )
 
--- a/spin/cmds/meson.py
+++ b/spin/cmds/meson.py
@@ -319,12 +319,19 @@ def build(
             )
         return
 
-    meson_args_setup = list(meson_args["setup"])
+    if isinstance(meson_args, tuple):
+        meson_args_ = {}
+        meson_args_["setup"] = meson_args
+        meson_args_["compile"] = tuple()
+        meson_args_["install"] = tuple()
+        meson_args = meson_args_
+
+    meson_args_setup = list(meson_args.get("setup", tuple()))
 
     if gcov:
         meson_args_setup = meson_args_setup + ["-Db_coverage=true"]
 
-    setup_cmd = _meson_cli() + ["setup", build_dir, "--prefix=/usr"] + meson_args_setup
+    setup_cmd = _meson_cli() + ["setup", abs_build_dir, "--prefix={}".format(abs_install_dir)] + meson_args_setup
 
     if clean:
         print(f"Removing `{build_dir}`")
@@ -351,30 +358,27 @@ def build(
 
         # Any other conditions that warrant a reconfigure?
 
-    meson_args_compile = list(meson_args["compile"])
+    meson_args_compile = list(meson_args.get("compile", tuple()))
     compile_flags = ["-v"] if verbose else []
     if "jobs" in meson_args:
         jobs = meson_args["jobs"]
-    compile_flags += ["-j", str(jobs)]
+    if jobs is not None:
+        compile_flags += ["-j", str(jobs)]
 
     p = _run(
-        _meson_cli() + ["compile"] + compile_flags + ["-C", build_dir] ,
+        _meson_cli() + ["compile"] + compile_flags + ["-C", abs_build_dir] + meson_args_compile,
         sys_exit=True,
         output=not quiet,
     )
 
-    meson_args_install = list(meson_args["install"])
+    meson_args_install = list(meson_args.get("install", tuple()))
     p = _run(
         _meson_cli()
         + [
             "install",
             "--only-changed",
             "-C",
-            build_dir,
-            "--destdir",
-            install_dir
-            if os.path.isabs(install_dir)
-            else os.path.relpath(abs_install_dir, abs_build_dir),
+            build_dir
         ] + meson_args_install,
         output=(not quiet) and verbose,
     )
--- a/spin/cmds/meson.py
+++ b/spin/cmds/meson.py
@@ -384,13 +384,7 @@ def build(
         ] + meson_args_install
 
     p = _run(
-         _meson_cli()
-         + [
-             "install",
-             "--only-changed",
-             "-C",
-             build_dir,
-             ] + meson_args_install,
+         cmd,
          output=(not quiet) and verbose,
      )
 
--- a/spin/cmds/meson.py
+++ b/spin/cmds/meson.py
@@ -1,6 +1,5 @@
 import contextlib
 import copy
-import datetime
 import json
 import os
 import re
@@ -8,8 +7,6 @@ import shutil
 import sys
 from enum import Enum
 from pathlib import Path
-import subprocess
-import time
 
 import click
 
--- a/spin/cmds/meson.py
+++ b/spin/cmds/meson.py
@@ -274,6 +274,8 @@ def build(
     quiet=False,
     build_dir=None,
     prefix=None,
+    meson_compile_args=tuple(),
+    meson_install_args=tuple(),
 ):
     """🔧 Build package with Meson/ninja
 
@@ -319,18 +321,7 @@ def build(
             )
         return
 
-    if isinstance(meson_args, tuple):
-        # for backwards compatibility because
-        # previously meson_args was a tuple.
-        # all the members of the meson_args tuple
-        # were passed to meson setup subcommand.
-        meson_args_ = {}
-        meson_args_["setup"] = meson_args
-        meson_args_["compile"] = tuple()
-        meson_args_["install"] = tuple()
-        meson_args = meson_args_
-
-    meson_args_setup = list(meson_args.get("setup", tuple()))
+    meson_args_setup = list(meson_args)
 
     if gcov:
         meson_args_setup = meson_args_setup + ["-Db_coverage=true"]
@@ -362,26 +353,24 @@ def build(
 
         # Any other conditions that warrant a reconfigure?
 
-    meson_args_compile = list(meson_args.get("compile", tuple()))
+    meson_compile_args = list(meson_compile_args)
     compile_flags = ["-v"] if verbose else []
-    if "jobs" in meson_args:
-        jobs = meson_args["jobs"]
     if jobs is not None:
         compile_flags += ["-j", str(jobs)]
 
     p = _run(
-        _meson_cli() + ["compile"] + compile_flags + ["-C", abs_build_dir] + meson_args_compile,
+        _meson_cli() + ["compile"] + compile_flags + ["-C", abs_build_dir] + meson_compile_args,
         sys_exit=True,
         output=not quiet,
     )
 
-    meson_args_install = list(meson_args.get("install", tuple()))
+    meson_install_args = list(meson_install_args)
     cmd = _meson_cli() + [
             "install",
             "--only-changed",
             "-C",
             build_dir
-        ] + meson_args_install
+        ] + meson_install_args
 
     p = _run(
          cmd,
--- a/spin/cmds/meson.py
+++ b/spin/cmds/meson.py
@@ -322,13 +322,6 @@ def build(
             )
         return
 
-    if isinstance(meson_args, tuple):
-        meson_args_ = {}
-        meson_args_["setup"] = meson_args
-        meson_args_["compile"] = tuple()
-        meson_args_["install"] = tuple()
-        meson_args = meson_args_
-
     meson_args_setup = list(meson_args.get("setup", tuple()))
 
     if gcov:
--- a/spin/cmds/meson.py
+++ b/spin/cmds/meson.py
@@ -992,7 +992,11 @@ def docs(
 
     make_bat_exists = (Path(doc_dir) / "make.bat").exists()
     make_cmd = "make.bat" if sys.platform == "win32" and make_bat_exists else "make"
-    _run([make_cmd, sphinx_target], cwd=doc_dir, replace=True)
+    cmds = [make_cmd, "-j", str(jobs), sphinx_target]
+    if jobs == "auto":
+        cmds = [make_cmd, "-j", sphinx_target]
+
+    _run(cmds, cwd=doc_dir, replace=True)
 
 
 @click.command()
--- a/spin/cmds/meson.py
+++ b/spin/cmds/meson.py
@@ -322,6 +322,17 @@ def build(
             )
         return
 
+    if isinstance(meson_args, tuple):
+        # for backwards compatibility because
+        # previously meson_args was a tuple.
+        # all the members of the meson_args tuple
+        # were passed to meson setup subcommand.
+        meson_args_ = {}
+        meson_args_["setup"] = meson_args
+        meson_args_["compile"] = tuple()
+        meson_args_["install"] = tuple()
+        meson_args = meson_args_
+
     meson_args_setup = list(meson_args.get("setup", tuple()))
 
     if gcov:
--- a/spin/cmds/meson.py
+++ b/spin/cmds/meson.py
@@ -1,5 +1,6 @@
 import contextlib
 import copy
+import datetime
 import json
 import os
 import re
@@ -7,6 +8,8 @@ import shutil
 import sys
 from enum import Enum
 from pathlib import Path
+import subprocess
+import time
 
 import click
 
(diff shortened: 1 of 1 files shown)


rgommers:
This does look like a useful improvement, thanks for working on it @czgdp1807. From the code I think I deduce (but didn't verify) that the changes aren't backwards compatible - is that right? If so, is it possible to just add the extra arguments at the end to preserve backwards compat?

czgdp1807:
> From the code I think I deduce (but didn't verify) that the changes aren't backwards compatible - is that right?

Yes. I will make it backwards compatible by tonight. I updated NumPy instead but that isn't the right approach because there are users of `spin` outside NumPy, SciPy. Closed the NumPy PR.

spin/cmds/meson.py
```diff
@@ -307,12 +310,12 @@ def build(
...
 
-    meson_args = list(meson_args)
+    meson_args_setup = list(meson_args.get("setup", tuple()))
 
     if gcov:
-        meson_args = meson_args + ["-Db_coverage=true"]
+        meson_args_setup = meson_args_setup + ["-Db_coverage=true"]
 
-    setup_cmd = _meson_cli() + ["setup", build_dir, "--prefix=/usr"] + meson_args
+    setup_cmd = _meson_cli() + ["setup", abs_build_dir, "--prefix={}".format(abs_install_dir)] + meson_args_setup
```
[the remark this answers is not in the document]
ngoldbaum: I think I'd prefer approach in my PR where it's possible to explicitly set `prefix`, I don't think that's possible here.

I'd also really like it if it were possible for downstream projects to adapt to changes in commands without explicitly updating their own spin code.

spin/cmds/meson.py
```diff
@@ -339,29 +342,72 @@ def build(
...
+                        elapsed = datetime.datetime.now() - start_time
+                        print(f"    ... installation in progress ({elapsed} "
+                                "elapsed)")
+                        last_blip = time.time()
+                        last_log_size = log_size
+
+            ret = p.wait()
+        except:  # noqa: E722
+            p.terminate()
+            raise
```
ngoldbaum, line 402: I'd be confused by this behavior on my personal machine I think. I'd rather see this sort of thing in project's CI config I think...

ngoldbaum:
> I updated NumPy instead but that isn't the right approach because there are users of spin outside NumPy, SciPy. Closed the NumPy PR.

I don't think NumPy *can* use this without updating NumPy's spin config to match. That at least was my experience adding a `prefix` argument, thus my opening https://github.com/scientific-python/spin/issues/242

czgdp1807:
One question to ask - What would be the right behaviour in case of NumPy? `spin build` and `spin test` should work perfectly? And `import numpy` should work without any issues?

ngoldbaum:
Sorry, I don't quite understand what you're asking.

Upstream spin changes shouldn't break dow stream projects, no. Ideally also (IMO) projects should be able to adopt changes to commands defined in spin itself by simply updating spin. Right now at least NumPy had to update spin, then update any commands NumPy overrides to get the new feature. Already NumPy's `spin build` is out of date and can't use the `gcov` argument that's defined in the upstream
command. The docstring for numpy's `spin build` is also out of date.

This is a shame because all NumPy really wants to do is add a single argument to `spin build`. IMO it should be possible to do that without updating the signature or docstring of NumPy's `build` function.

czgdp1807:
I see. Let me reword my question. I just want to check whether my changes here break anything in NumPy's build process. If `spin build` and `spin test` work fine then can I assume that my changes haven't broken anything in NumPy.

rgommers:
> Let me reword my question. I just want to check whether my changes here break anything in NumPy's build process. If `spin build` and `spin test` work fine then can I assume that my changes haven't broken anything in NumPy.

Yes, that should be the case. After this PR is updated to remove backwards-incompatible changes, it should also be straightforward for the code reviewers here to see that nothing could have broken. All you want is a new way to pass on extra flags to `meson compile` and `meson install`; if `numpy` doesn't yet use that new feature, nothing should be breaking.

czgdp1807:
`spin build` and `spin test` works perfectly in NumPy with this PR. So I think nothing breaks in NumPy because of the updates I made in this PR.

czgdp1807:
> Update - The support for `spin bench` is complete in SciPy (here) and in `spin`. `spin bench` and `python dev.py bench` work in the same manner.

stefanv:
Thank you so much @czgdp1807. I'm sorry that I haven't reviewed yet; please feel free to bug me if you don't hear from me by Tuesday.

czgdp1807:
So I discussed with @rgommers and he told me that `bench` shouldn't be here. That's 100% correct. I should have checked how NumPy implemented `spin bench` and then should have taken things from there. Anyways, now I will make this PR ready for reviews. What I need to do here is the following,

- [ ] Drop the `bench` changes from here.
- [ ] Explain the rest of the changes.

Once these two are done, it will be ready for review.

spin/cmds/meson.py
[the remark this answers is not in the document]
czgdp1807: Makes sense to me. Basically the changes I made work for both SciPy and NumPy as well. The changes in `main` branch of `spin` work only with NumPy and not with SciPy. Your change is better (it is backwards compatible), however since `--prefix=/usr` doesn't work with SciPy, so we will have to pass `--prefix=<abs_install_dir>` in SciPy project. 

Basically the choice to be made here is,

1. Option 1 - Follow https://github.com/scientific-python/spin/pull/241/files and pass `--prefix=<abs_build_dir>`.
2. Option 2 - Follow this PR (once it's merged in `main` branch of `spin`). Then NumPy and SciPy will work fine as is. However, its hard to say if other projects using `spin` will break or not.

I am fine with Option 1. That way only SciPy interface will need changes and no other project using spin will break.

cc: @rgommers @stefanv

spin/cmds/meson.py
```diff
@@ -351,30 +362,37 @@ def build(
...
-            build_dir,
-            "--destdir",
-            install_dir
-            if os.path.isabs(install_dir)
-            else os.path.relpath(abs_install_dir, abs_build_dir),
-        ],
-        output=(not quiet) and verbose,
-    )
+            build_dir
+        ] + meson_args_install
```
czgdp1807, line 384: This also may or may not be backwards compatible. The change made here is removal of `--destdir`. I think with this the `--prefix` will be used as installation directory.

Let me know if I am mis-interpreting things.

spin/cmds/meson.py
```diff
@@ -351,30 +362,37 @@ def build(
...
+    p = _run(
+         _meson_cli()
+         + [
+             "install",
+             "--only-changed",
+             "-C",
+             build_dir,
+             ] + meson_args_install,
+         output=(not quiet) and verbose,
+     )
```
czgdp1807, line 395: This is harmless and backwards compatible change IMO.

spin/cmds/meson.py
```diff
@@ -946,7 +964,11 @@ def docs(
 
     make_bat_exists = (Path(doc_dir) / "make.bat").exists()
     make_cmd = "make.bat" if sys.platform == "win32" and make_bat_exists else "make"
-    _run([make_cmd, sphinx_target], cwd=doc_dir, replace=True)
+    cmds = [make_cmd, "-j", str(jobs), sphinx_target]
+    if jobs == "auto":
+        cmds = [make_cmd, "-j", sphinx_target]
+
+    _run(cmds, cwd=doc_dir, replace=True)
```
czgdp1807, line 971: This is also harmless and backwards compatible IMO. Just adds support for parallel build of docs.

czgdp1807 (commented):
So this PR is ready for review and I have explained the changes to the best of my abilities. Note https://github.com/scientific-python/spin/pull/238#discussion_r1834424443 too other than the below comments.

cc: @stefanv

spin/cmds/meson.py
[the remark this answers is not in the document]
rgommers: SciPy has the same constraints as NumPy. It doesn't matter what `--prefix` is set to, I'd expect things to work for both project. The actual install location is controlled by `--destdir`; `--prefix` is turned into a relative path under `destdir`. Look in `numpy/build-install/` to see what I mean (the next subdir there is `usr`).

rgommers:
> Drop the `bench` changes from here.

👍🏼. Can you edit the PR title as well to reflect this?

spin/cmds/meson.py
rgommers: I think you don't need these changes. What goes wrong if you don't touch this code at all? If the only thing that changes is the install path going from this:
```
build-install/lib/python3.12/site-packages/scipy/
```
to:
```
build-install/usr/lib/python3.12/site-packages/scipy/
```
then that is fine, that's a completely harmless change.

czgdp1807: Actually I we follow Option 1 from https://github.com/scientific-python/spin/pull/238#discussion_r1834424443 and undo this change then from my last experience, `spin test` didn't work correctly. Actually I made the changes here to match with the series of commands executed by `python dev.py build`. Then everything works in SciPy (also with NumPy). I can try undoing this changes and let you know if SciPy still works.

czgdp1807: Here is the error if I undo my changes,

```zsh
(scipy-dev) 20:11:53:~/Quansight/scipy % spin test -s stats
$ meson compile -j 10 -C /Users/czgdp1807/Quansight/scipy/build
INFO: autodetecting backend as ninja
INFO: calculating backend command to run: /Users/czgdp1807/mambaforge/envs/scipy-dev/bin/ninja -C /Users/czgdp1807/Quansight/scipy/build -j 10
ninja: Entering directory `/Users/czgdp1807/Quansight/scipy/build'
[1/1] Generating scipy/generate-version with a custom command
$ meson install --only-changed -C build --destdir ../build-install --tags=runtime,python-runtime,tests,devel
$ export PYTHONPATH="/Users/czgdp1807/Quansight/scipy/build-install/usr/lib/python3.12/site-packages"
$ /Users/czgdp1807/mambaforge/envs/scipy-dev/bin/python3.12 -P -c 'import scipy'
$ cd /Users/czgdp1807/Quansight/scipy/build-install/usr/lib/python3.12/site-packages
$ /Users/czgdp1807/mambaforge/envs/scipy-dev/bin/python3.12 -P -m pytest --import-mode=importlib --pyargs scipy.stats -m 'not slow'
===================================================== test session starts ======================================================
platform darwin -- Python 3.12.6, pytest-8.3.3, pluggy-1.5.0
rootdir: /Users/czgdp1807/Quansight/scipy
configfile: pytest.ini
plugins: cov-5.0.0, anyio-4.4.0, hypothesis-6.112.1, timeout-2.3.1, xdist-3.6.1
collected 0 items / 35 errors                                                                                                  

============================================================ ERRORS ============================================================
___________________ ERROR collecting build-install/usr/lib/python3.12/site-packages/scipy/stats/_bws_test.py ___________________
ImportError while importing test module '/Users/czgdp1807/Quansight/scipy/build-install/usr/lib/python3.12/site-packages/scipy/stats/_bws_test.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
scipy/stats/__init__.py:614: in <module>
    from ._stats_py import *
scipy/stats/_stats_py.py:39: in <module>
    from scipy.spatial import distance_matrix
scipy/spatial/__init__.py:116: in <module>
    from ._geometric_slerp import geometric_slerp
scipy/spatial/_geometric_slerp.py:9: in <module>
    from scipy.spatial.distance import euclidean
scipy/spatial/distance.py:121: in <module>
    from ..special import rel_entr
scipy/special/__init__.py:816: in <module>
    from . import _ufuncs
E   ImportError: dlopen(/Users/czgdp1807/Quansight/scipy/build-install/usr/lib/python3.12/site-packages/scipy/special/_ufuncs.cpython-312-darwin.so, 0x0002): Library not loaded: /usr/lib/python3.12/site-packages/scipy/special/libsf_error_state.dylib
E     Referenced from: <C40E1CEC-A9D0-39F7-8466-1ABD39B3EF5D> /Users/czgdp1807/Quansight/scipy/build-install/usr/lib/python3.12/site-packages/scipy/special/_ufuncs.cpython-312-darwin.so
E     Reason: tried: '/usr/lib/python3.12/site-packages/scipy/special/libsf_error_state.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/usr/lib/python3.12/site-packages/scipy/special/libsf_error_state.dylib' (no such file), '/usr/lib/python3.12/site-packages/scipy/special/libsf_error_state.dylib' (no such file, not in dyld cache)
```

rgommers: The shared library bites again:) No worries, I'm pretty sure I can fix that. I'll try to have a look later today.

To help me here, could you point me to a different `spin` branch you have that allows me to reproduce this problem from a `spin test` in a clean repo?

czgdp1807: Yes. `spin_upd_1` (the branch which contains the changes I reverted from here). Link - https://github.com/czgdp1807/spin/tree/spin_upd_1

rgommers: It's the same issue as is being discussed in https://github.com/numpy/numpy/pull/27371, it has nothing to do with the shared library but with a breaking change in `spin` for test discovery that happened in a recent release.

E.g., for `spin test -s stats`:
```
ImportError while importing test module '/home/rgommers/code/scipy/build-install/usr/lib/python3.12/site-packages/scipy/stats/_page_trend_test.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
scipy/stats/_page_trend_test.py:4: in <module>
    from ._continuous_distns import norm
E   ModuleNotFoundError: No module named 'build-install.usr.lib.python3.12.site-packages.scipy.stats._continuous_distns'; 'build-install.usr.lib.python3.12.site-packages.scipy.stats' is not a package
...
```

And for `spin test -s cluster`:
```
ImportError while importing test module '/home/rgommers/code/scipy/build-install/usr/lib/python3.12/site-packages/scipy/cluster/tests/test_hierarchy.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
scipy/cluster/tests/test_hierarchy.py:54: in <module>
    from . import hierarchy_test_data
E   ImportError: cannot import name 'hierarchy_test_data' from 'build-install.usr.lib.python3.12.site-packages.scipy.cluster.tests' (unknown location)
```

Forbidding relative imports in tests is not great. SciPy has quite a few of those.

spin/cmds/meson.py
```diff
@@ -319,12 +319,23 @@ def build(
...
-    meson_args = list(meson_args)
+    if isinstance(meson_args, tuple):
+        # for backwards compatibility because
+        # previously meson_args was a tuple.
+        # all the members of the meson_args tuple
+        # were passed to meson setup subcommand.
+        meson_args_ = {}
+        meson_args_["setup"] = meson_args
+        meson_args_["compile"] = tuple()
+        meson_args_["install"] = tuple()
```
stefanv, line 330: I don't see these being used below, so do we need the complication of splitting meson_args into three parts?

stefanv: OK, thanks for clarifying. I expected to see some changes to the click parameters to allow specifying such separate sets of settings, but I don't?

czgdp1807: Yes. It wasn’t needed.

stefanv: So what cli usage will correctly populate this parameter?

czgdp1807: In SciPy, the code in `spin/.cmds.py`, we add code to figure out which CLI argument should go into which subcommand. If you don't do it, then the code above will be activated and everything will be passed to `meson setup` subcommand.

czgdp1807: Let me explain more. So in SciPy, we have some SciPy specific build options which are passed via CLI. Now these options are mapped to `meson` arguments. `meson` is called here in `spin`. Some SciPy build options are forwarded to `meson setup` subcommand, some to `meson compile` and some to `meson install`. That's why to need to figure out which SciPy build option will go into `setup`, which ones will go into `compile` and which ones will go into `install`. Therefore this segregation can only be done in SciPy. Same can be happen with any other project, they can define their own build options which will map to some `meson` arguments. They will segregate those options into `setup`, `compile` and `install`. The `spin` code here in this PR will generate `meson` sub-commands and attach those forwarded options to those subcommands.

Let me know if I misinterpreted anything.

czgdp1807: See https://github.com/scipy/scipy/blob/20f8f5eca1f90c49ffe1133c5af3b22cba31fe65/.spin/cmds.py#L345-L406 to see how the segregation of SciPy build options into meson subcommads is happening.

czgdp1807: Basically it's about how `meson` is being called in `spin`. If I want to pass a meson argument to `meson install`, then `spin` won't do it. All the `meson_args` are attached to `meson setup` subcommand.

Let me know if I am wrong here.

czgdp1807: > Then after updating scipy's spin config to use the new extension mechanism added in #248 all projects using spin and meson will get the feature "for free".

I am not sure if the issue I am solving here is linked to #248. I will go through it and see if it can be used here as well.

stefanv: Is there a reason other projects would not benefit from being able to specify those three settings separately? If they would, we could add flags to allow their specification?

If not, I think best to split this into three keyword arguments. That would still allow SciPy to do what it needs, but it's less obscure from the spin perspective (the implicit dictionary structure being used now).

rgommers: > Is there a reason other projects would not benefit from being able to specify those three settings separately? If they would, we could add flags to allow their specification?

Yes, being able to pass arguments to `meson compile` and `meson install` is potentially useful to any project that uses Meson.

This PR in its current form does allow that. If it's easier/nicer to add 3 new keywords rather than mash it into `meson_args`, then sure let's do that.

czgdp1807: Made the changes as discussed here.

spin/cmds/meson.py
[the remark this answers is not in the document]
rgommers: Another `--prefix` patch landed in `main` already. Did you try dropping this change @czgdp1807?

spin/cmds/meson.py
```diff
@@ -351,30 +353,29 @@ def build(
...
 
+    meson_compile_args = list(meson_compile_args)
     compile_flags = ["-v"] if verbose else []
-    if jobs:
+    if jobs is not None:
         compile_flags += ["-j", str(jobs)]
 
     p = _run(
-        _meson_cli() + ["compile"] + compile_flags + ["-C", build_dir],
+        _meson_cli() + ["compile"] + compile_flags + ["-C", abs_build_dir] + meson_compile_args,
```
rgommers, line 362: An absolute directory is certainly not needed here, `-C` takes a relpath

spin/cmds/meson.py
rgommers: I'm not sure this is backwards compatible, parallel Sphinx builds are breaking regularly (example: https://github.com/scipy/scipy/issues/18300). From the diff I can't quite tell if the default behavior is switched, but the code looks not quite right to me, since if `jobs == 'auto'` then `-j` gets added twice (in lines 950 and 952).

It's also unrelated to the topic of this PR, so I suggest removing the changes. We need this PR to get merged and in a release, so we can move on with `spin` adoption in SciPy - it's important to keep this focused on the issue at hand. If we need other changes unrelated to passing `meson compile` and `meson install`, please put them in a separate PR (or more, one per topic would be great unless things are completely trivial).

spin/cmds/meson.py
rgommers: This seems to work smoothly in spin 0.13 now. I think the comment can be resolved, and no more changes should be needed (🤞🏼).

spin/cmds/meson.py
czgdp1807: Sure. No worries. I added this in the beginning because `dev.py` has parallel building of docs available. I think we should remove it from `dev.py` also.

czgdp1807:
I will open up a new PR for keyword change. Rest will be kept here.

czgdp1807:
Seems like several improvements were made. The only one left was keyword arguments for `meson compile` and `meson install`. So I am closing this PR in favour of https://github.com/scientific-python/spin/pull/256

czgdp1807:
Context - This PR was opened at a time when `spin` didn't have features which I required to make it a drop in replacement of `scipy/dev.py`. However, at the current stage, everything is there except one feature. It is implemented in #256 and waiting for merge.

PR #241, opened 2024-10-16
Allow overriding --prefix in spin build

### LINKED ISSUE #240: "spin build" errors on windows
When I try to use spin to build numpy on windows, I get:

```
meson.build:1:0: ERROR: prefix balue '/usr' must be an absolute path
```

It looks like this is hard-coded in the `build` command. I've never had an issue with it before so no idea why it's problematic now.

ngoldbaum:
Fixes #240

This lets users override --prefix. Right now it's hard-coded, which leads to breakage on windows sometimes.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -38,24 +38,3 @@ jobs:
       - name: Tests PyTest
         run: |
           pipx run nox --forcecolor -s test
-
-  test-uv-python:
-    strategy:
-      matrix:
-        python_version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13", "3.13t"]
-        os: [ubuntu-latest, windows-latest, macos-latest]
-    runs-on: ${{ matrix.os }}
-    steps:
-      - uses: actions/checkout@v4
-      - uses: astral-sh/setup-uv@v3
-      - run: |
-          uv python install ${{ matrix.python_version }}
-          uv venv --python ${{ matrix.python_version }}
-      - name: Install system dependencies
-        if: matrix.os == 'ubuntu-latest'
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y gdb lcov
-      - name: Tests PyTest
-        run: |
-          pipx run nox --forcecolor -s test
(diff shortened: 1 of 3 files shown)


ngoldbaum:
@stefanv I think your existing windows tests do run on powershell. You can see what shell ana ctions step is running by expanding the input details.

image


So I think the issue isn't powershell so much as how python is set up. I can trigger the problem using a python installed via micromamba and uv on Windows.

stefanv:
We don't actually install to `/usr`, though, that is the prefix used under the `build-install` dir. I think that is in place to make some of the checking work, by having the path names be predictable.

The tests seem happy, though, so perhaps you're right!

stefanv:
> python installed via micromamba and uv on Windows

This is very helpful; I could make a matrix entry for that.

ngoldbaum:
For uv you can do something like this on github actions:

```
    - uses: astral-sh/setup-uv@v3
    - run: |
        uv python install ${{ matrix.version }}
        uv venv --python ${{ matrix.version }}
    - run: |
        .venv\Scripts\activate
        uv pip install some_dependency
        python -m pytest
```

ngoldbaum:
Darn, it looks like simply installing python from uv wasn't enough to trigger it in the spin tests. I'm not sure how to write a test using the existing framework to trigger what I'm seeing with NumPy.

I guess I could add a CI job that just builds NumPy?

I'll leave it up to you, please feel free to push to this PR if you want to adjust things.

stefanv:
Thank you so much, Nathan. I'm out until next week, but will take a look at it then.

ngoldbaum:
OK, I've added a new `--prefix` option to `meson.build`. If I update NumPy's `build` command to match meson's everything works and I'm able to use `spin build --prefix` with numpy. Also added a test.

I also ran into #242, I think it's too easy for commands that projects want to override to go out of sync with upstream improvements.

ngoldbaum:
I also just checked on the Windows system I originally had a problem on - changing the default prefix to `"C:\\"` allows me to build on Windows.

ngoldbaum:
Hi, any chance we can get this reviewed? I tried again today to set up Windows tests for NumPy using `spin` on the free-threaded build using an alternate way to install Python and I hit this issue again, so setting up tests on Windows is currently blocked on this.

stefanv (approved):

PR #246, opened 2024-11-08
Let subprocess launched by run handle its own sigint

### LINKED ISSUE #235: spin catches SIGINT and send SIGTERM to subprocess
Seems like `spin` is sending a `SIGTERM` (or `SIGKILL`?)signal to the underlying process when it receives a `SIGINT`, which seems odd when working in a `spin run python -i /path/to/script.py` kinda command.

Shouldn't spin forward a `SIGINT` instead? while running the process in `python`, I wouldn't expect `CTRL+C` to exit the process really.

stefanv:
Otherwise, the subprocess is aborted by the parent process. This change makes `spin run python -i ...` usable.

Closes #235

/cc @adrinjalali

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,6 @@ dependencies = [
   "click",
   "tomli; python_version < '3.11'",
   "colorama; platform_system == 'Windows'",
-  "pywin32; platform_system == 'Windows'",
   "importlib_metadata >= 7"
 ]
 dynamic = ['version']
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,6 +26,7 @@ dependencies = [
   "click",
   "tomli; python_version < '3.11'",
   "colorama; platform_system == 'Windows'",
+  "pywin32; platform_system == 'Windows'",
   "importlib_metadata >= 7"
 ]
 dynamic = ['version']
--- a/spin/cmds/meson.py
+++ b/spin/cmds/meson.py
@@ -5,6 +5,7 @@ import os
 import re
 import shutil
 import signal
+import subprocess
 import sys
 from enum import Enum
 from pathlib import Path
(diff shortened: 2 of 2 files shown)


adrinjalali:
This simply ignores the signal though. We'd want to delegate it to the subprocess.

With this PR, in numpy's repo, I get:

```py
$ spin run python
$ /tmp/spin/bin/python3 vendored-meson/meson/meson.py compile -C build
$ /tmp/spin/bin/python3 vendored-meson/meson/meson.py install --only-changed -C build --destdir ../build-install
Python 3.12.7 (main, Oct  1 2024, 11:15:50) [GCC 14.2.1 20240910] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import time
>>> time.sleep(10)
^C^C^C>>> 
```

while with a simple python process you'd get:

```py
$ python
Python 3.12.7 (main, Oct  1 2024, 11:15:50) [GCC 14.2.1 20240910] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import time
>>> time.sleep(10)
^CTraceback (most recent call last):
  File "<stdin>", line 1, in <module>
KeyboardInterrupt
>>> 
```

stefanv:
OK, will dig a bit further.

stefanv:
Hopefully got it this time!

adrinjalali:
Tests supposed to be failing?

stefanv:
> Tests supposed to be failing?

No, it's on Windows only, will see why; Windows probably doesn't support preexec_fn.

adrinjalali:
But I can confirm this fixes the issue nicely in linux at least.

stefanv:
Alternative approach to try:

```python
try:
    process.wait()
except KeyboardInterrupt:
    print("Terminating subprocess...")
    parent = psutil.Process(process.pid)
    for child in parent.children(recursive=True):
        child.kill()
    process.wait()
```

adrinjalali:
I think I like this as is now, we shouldn't decide for the process that sigint should kill it. If the process wants to do a dance on that signal, it's free to do so 😁

stefanv:
I just need to fire up my Windows test machine, then I can see if it works on that platform as well.

stefanv:
Tested on Windows, and seems like the behavior is as intended without any changes, so reverted.

PR #247, opened 2024-11-08
Improve source layout test discovery

stefanv:
This should allow us to move forward with https://github.com/numpy/numpy/pull/27371 without breaking behavior for `src/tests` layout.


ngoldbaum:
Thanks so much for working on this! Let me know if you need anything from me on the NumPy side of things to get this in and update NumPy's CI to spin 0.13 afterwards.

PR #248, opened 2024-11-13
Add mechanism for extending existing spin commands

### LINKED ISSUE #242: It should be easier to extend commands provided by spin
NumPy [defines a `spin build` command](https://github.com/numpy/numpy/blob/dc862b43e3bca09dcf5c235c46b2c7573e2e4b80/.spin/cmds.py#L96), which has fallen out of sync from the upstream `meson.build` command it uses. Reading the click docs, it seems like the pattern NumPy is using with `ctx.forward` is frowned upon.

I think it would be better if spin provided factories for click commands to make it easier for projects to extend upstream commands without having to effectively re-implement their API.

See e.g. https://stackoverflow.com/questions/55584012/python-click-dependent-options-on-another-option.

stefanv:
Closes #242

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,11 +207,11 @@ For this purpose, we provide the `spin.util.extend_cmd` decorator.
 Here, we show how to add a `--extra` flag to the existing `build` function:
 
 ```python
-from spin.cmds import meson
+import spin
 
 
 @click.option("-e", "--extra", help="Extra test flag")
-@util.extend_command(spin.cmds.meson.build)
+@spin.util.extend_command(spin.cmds.meson.build)
 def build_extend(*, parent_callback, extra=None, **kwargs):
     """
     This version of build also provides the EXTRA flag, that can be used
--- a/README.md
+++ b/README.md
@@ -233,6 +233,8 @@ The matching entry in `pyproject.toml` is:
 The `extend_cmd` decorator also accepts a `doc` argument, for setting the new command's `--help` description.
 The function documentation ("This version of build...") is also appended.
 
+Finally, `remove_args` is a tuple of arguments that are not inherited from the original command.
+
 ### Advanced: override Meson CLI
 
 Some packages use a vendored version of Meson. The path to a custom
--- a/README.md
+++ b/README.md
@@ -201,52 +201,37 @@ click options or function keywords:
 
 ### Advanced: adding arguments to built-in commands
 
-Instead of rewriting a command from scratch, a project may want to add a flag to a built-in `spin` command, or perhaps do some pre- or post-processing.
-For this, we have to use an internal Click concept called a [context](https://click.palletsprojects.com/en/8.1.x/complex/#contexts).
-Fortunately, we don't need to know anything about contexts other than that they allow us to execute commands within commands.
+Instead of rewriting a command from scratch, a project may simply want to add a flag to an existing `spin` command, or perhaps do some pre- or post-processing.
+For this purpose, we provide the `spin.util.extend_cmd` decorator.
 
-We proceed by duplicating the function header of the existing command, and adding our own flag:
+Here, we show how to add a `--extra` flag to the existing `build` function:
 
 ```python
 from spin.cmds import meson
 
-# Take this from the built-in implementation, in `spin.cmds.meson.build`:
 
-
-@click.command()
-@click.argument("meson_args", nargs=-1)
-@click.option("-j", "--jobs", help="Number of parallel tasks to launch", type=int)
-@click.option("--clean", is_flag=True, help="Clean build directory before build")
-@click.option(
-    "-v", "--verbose", is_flag=True, help="Print all build output, even installation"
-)
-
-# This is our new option
-@click.option("--custom-arg/--no-custom-arg")
-
-# This tells spin that we will need a context, which we
-# can use to invoke the built-in command
-@click.pass_context
-
-# This is the original function signature, plus our new flag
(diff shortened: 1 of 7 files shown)


stefanv:
@ngoldbaum Please take a look and see if [this API](https://github.com/scientific-python/spin/blob/9aac7130dfe51d169cbf005662512e19ead1d8f9/README.md#advanced-adding-arguments-to-built-in-commands) covers your use cases.

README.md
```diff
@@ -201,52 +201,37 @@ click options or function keywords:
...
-We proceed by duplicating the function header of the existing command, and adding our own flag:
+Here, we show how to add a `--extra` flag to the existing `build` function:
 
 ```python
 from spin.cmds import meson
 
-# Take this from the built-in implementation, in `spin.cmds.meson.build`:
 
+@click.option("-e", "--extra", help="Extra test flag")
+@util.extend_command(spin.cmds.meson.build)
```
ngoldbaum, line 214: maybe make this `spin.util.extend_command` or an explicit import to this example?

ngoldbaum (commented):
Neat! I'll try to see if this solves the issues in NumPy soon. Just a quick drive-by comment I noticed looking at the docs.

stefanv:
I also added argument removal, which is a feature you previously requested.

ngoldbaum:
This is awesome! I love deleting code and ending up with a better result:

<details>

```
goldbaum at Nathans-MBP in ~/Documents/numpy on main!
± git diff
diff --git a/.spin/cmds.py b/.spin/cmds.py
index fea3565a5e..12ee17878c 100644
--- a/.spin/cmds.py
+++ b/.spin/cmds.py
@@ -1,8 +1,9 @@
+import importlib
 import os
-import shutil
 import pathlib
-import importlib
+import shutil
 import subprocess
+import sys

 import click
 from spin import util
@@ -72,46 +73,18 @@ def changelog(ctx, token, revision_range):
         )


-@click.command()
-@click.option(
-    "-j", "--jobs",
-    help="Number of parallel tasks to launch",
-    type=int
-)
-@click.option(
-    "--clean", is_flag=True,
-    help="Clean build directory before build"
-)
-@click.option(
-    "-v", "--verbose", is_flag=True,
-    help="Print all build output, even installation"
-)
 @click.option(
     "--with-scipy-openblas", type=click.Choice(["32", "64"]),
     default=None,
     help="Build with pre-installed scipy-openblas32 or scipy-openblas64 wheel"
 )
-@click.argument("meson_args", nargs=-1)
-@click.pass_context
-def build(ctx, meson_args, with_scipy_openblas, jobs=None, clean=False, verbose=False, quiet=False, *args, **kwargs):
-    """🔧 Build package with Meson/ninja and install
-
-    MESON_ARGS are passed through e.g.:
-
-    spin build -- -Dpkg_config_path=/lib64/pkgconfig
-
-    The package is installed to build-install
-
-    By default builds for release, to be able to use a debugger set CFLAGS
-    appropriately. For example, for linux use
-
-    CFLAGS="-O0 -g" spin build
-    """
-    # XXX keep in sync with upstream build
+@util.extend_command(meson.build)
+def build(
+        *, parent_callback, with_scipy_openblas=None, **kwargs
+):
     if with_scipy_openblas:
         _config_openblas(with_scipy_openblas)
-    ctx.params.pop("with_scipy_openblas", None)
-    ctx.forward(meson.build)
+    parent_callback(**kwargs)


 @click.command()

goldbaum at Nathans-MBP in ~/Documents/numpy on main!
± spin build --help
Usage: spin build [OPTIONS] [MESON_ARGS]...

  🔧 Build package with Meson/ninja

  The package is installed to `build-install` (unless a different build directory is specified with `-C`).

  MESON_ARGS are passed through e.g.:

    spin build -- -Dpkg_config_path=/lib64/pkgconfig

  By default meson-python does release builds. To be able to use a debugger, tell meson to build in debug mode:

    spin build -- -Dbuildtype=debug

  or set CFLAGS appropriately:

    CFLAGS="-O0 -g" spin build

  Build into a different build/build-install directory via the `-C/--build-dir` flag:

    spin build -C build-for-feature-x

  This feature is useful in combination with a shell alias, e.g.:

    $ alias spin-clang="SPIN_BUILD_DIR=build-clang CC=clang spin"

  Which can then be used to build (`spin-clang build`), to test (`spin-clang test ...`), etc.

Options:
  -j, --jobs INTEGER             Number of parallel tasks to launch
  --clean                        Clean build directory before build
  -v, --verbose                  Print detailed build and installation output
  --gcov                         Enable C code coverage using `gcov`. Use `spin test --gcov` to generate reports.
  --prefix TEXT                  The build prefix, passed directly to meson.  [default: /usr]
  -C, --build-dir TEXT           Meson build directory; package is installed into './{build-dir}-install'.  [env var: SPIN_BUILD_DIR; default: build]
  --with-scipy-openblas [32|64]  Build with pre-installed scipy-openblas32 or scipy-openblas64 wheel
  --help                         Show this message and exit.
```

</details>

I guess I could leave the old code in an if statement to support old versions of spin. Do you think that's worth doing in NumPy's spin config? Or should we just require spin 0.13 and substantially slim down our `.spin/cmds.py`?

stefanv:
Nice :)

I would just do a version check on `spin`, and provide a helpful error message / instruction.

stefanv:
If this works well for you, I'm inclined to merge this and make a release.

ngoldbaum:
That would be great. I'd also appreciate it if you could handle updating NumPy's spin config in the context of your open PR. I'm also happy to take that over if you don't think you'll have time for a while.

Issue #176, opened 2024-04-04
spin build should fix the rpath for shared libaries


oscarbenjamin:
I'm migrating python-flint to meson-python and testing out spin as a frontend (https://github.com/flintlib/python-flint/pull/129).

The setup that I use with python-flint is to build libraries with `--prefix=$(pwd)/.local` so that I have isolated builds of the C dependencies inside the python-flint development checkout.

When building wheels with meson-python I can set `PKG_CONFIG_PATH` to find the libraries and then meson-python fixes the rpath:
https://github.com/mesonbuild/meson-python/blob/c9d1c2dafb54692ad56dcda1bc67fd52b83b29c0/mesonpy/__init__.py#L431-L440

This means that I can do e.g.
```console
$ PKG_CONFIG_PATH=$(pwd)/.local/lib/pkgconfig spin install
 $ pip install --no-build-isolation --editable .
 ...
Successfully installed python-flint-0.6.0
$ python -m flint.test -qt
Running tests...
flint.test: all 49 tests passed!
----------------------------------------
OK: Your installation of python-flint seems to be working just fine!
----------------------------------------
```
However `spin build` does not fix the rpath and so `spin test` does not work unless I set `LD_LIBRARY_PATH`:
```console
$ PKG_CONFIG_PATH=$(pwd)/.local/lib/pkgconfig spin build --clean
...
[90/90] Linking target src/flint/functions/showgood.cpython-312-x86_64-linux-gnu.so
 $ meson install --only-changed -C build --destdir ../build-install
$ spin test
Invoking `build` prior to running tests:
 $ meson compile -C build
INFO: autodetecting backend as ninja
INFO: calculating backend command to run: /usr/bin/ninja -C /home/oscar/current/active/python-flint/build
ninja: Entering directory `/home/oscar/current/active/python-flint/build'
ninja: no work to do.
 $ meson install --only-changed -C build --destdir ../build-install
 $ export PYTHONPATH="/home/oscar/current/active/python-flint/build-install/usr/lib/python3.12/site-packages"
 $ /home/oscar/.pyenv/versions/3.12.0/envs/python-flint-3.12/bin/python3.12 -P -c 'import flint'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/oscar/current/active/python-flint/build-install/usr/lib/python3.12/site-packages/flint/__init__.py", line 1, in <module>
    from .pyflint import *
  File "pyflint.pyx", line 1, in init flint.pyflint
[...]

PR #257, opened 2024-12-05
ENH: support shared libraries inside packages

rgommers:
This drops the use of staged installs with `--destdir` in favor of using `--prefix` for the install location. The problem is that RPATHs in extension modules end up pointing to what Meson thinks is the final install location (i.e., inside the prefix) rather than to the staging directory.

`--destdir` seems meant for packaging, as an actual staging area, while for `spin` the final install directory is `build-install` (by default), there is no intent to later put this package into `/usr` or `C:\`. Hence using `--prefix` seems like the correct thing to do.

The one test change here is to a test that was incorrect. `meson setup --prefix` expects an absolute path, and `/foobar` isn't a path that exists or can be created.

Addresses the issue discussed in gh-238 - the `spin build` behavior before this change cannot be made to work for SciPy, because the internal shared library in `scipy.special` keeps breaking. It should also address the problem discussed in issue gh-176.

It may be my fault for suggesting the use of `--destdir` early on in `spin`'s life - I think I understand the tradeoffs a bit better now.

@stefanv question about testing - do you want me to extend `example_pkg` with a shared library? It will look like `sharedlib-in-package/mypkg/sub/*` in https://github.com/mesonbuild/meson-python/pull/700. It's a little bit of extra complexity, but it should be pretty robust since it mirrors what SciPy has internally.


czgdp1807 (approved):
Yes. This is exactly what we need. I think it is a non-breaking change. I did this change in #238 and it worked with NumPy as well. 

May be we can accept `––destdir` as an optional argument and pass it only when available. That way if someone needs it they can pass it via `spin build ––destdir`. In simple words, its just like what we did for `––prefix` in `spin`.

rgommers:
> May be we can accept `––destdir` as an optional argument and pass it only when available. That way if someone needs it they can pass it via `spin build ––destdir`.

I don't think anyone needs it - I'd avoid adding it unless there's a real-world use case. `spin` by design I think has one install directory per build directory, and that is it.

stefanv:
Good catch, thanks @rgommers! Happy to move to `--prefix`, I don't think that will negatively impact any other project. And, yes, no problem adding a minimally viable shared library to `example_pkg`.

rgommers:
I can't quite tell what's up with the editable install failures on Windows. Will need to switch to a Windows machine to see what's up - could be something trivial, but if it's not then it's probably relevant for SciPy as well when we switch to `spin`.

stefanv:
Windows is very finicky; I have a test machine here for that purpose, and can help debug if that'd be helpful.

rgommers:
> Windows is very finicky; I have a test machine here for that purpose, and can help debug if that'd be helpful.

Thanks. I'll try to have a look at it in the coming days, but if you have time to look at it that may be helpful. I haven't quite followed along with how editable install support in `spin` works.

stefanv:
`ImportError: DLL load failed while importing _core: The specified module could not be found.`

stefanv:
What is the status of this, now that SciPy is heading in a different direction?

stefanv:
Closing, since I don't think this is being actively pursued; please feel free to re-open if that's incorrect.

rgommers:
We removed the shared library inside SciPy, so this was no longer blocking. I think the change is still correct, but the test case wasn't fully working on Windows and it takes too much time to debug at the moment. So fine to close, we can revisit if and when the priority goes back up.

Thanks for the help!

PR #271, opened 2025-03-17
On run with replace, warn if executable not found

### LINKED ISSUE #270: An internal error has occurred when ipython not installed
I get this error from `spin ipython` when `ipython` is not installed:
```console
$ spin ipython
...
 $ ipython --ignore-cwd

Traceback (most recent call last):
  File "<frozen os>", line 611, in execvp
  File "<frozen os>", line 653, in _execvpe
  File "<frozen os>", line 644, in _execvpe
FileNotFoundError: [Errno 2] No such file or directory

An internal error has occurred. Please file a bug report at

  https://github.com/scientific-python/spin

including the above traceback and the following information:

  spin: 0.13, package: python-flint

Aborting.
```
Obviously it can't run ipython if I don't have ipython installed but the error message could be improved to at least mention ipython rather than suggesting to open a bug report here.

stefanv:
Closes #270

### DIFF
--- a/spin/cmds/meson.py
+++ b/spin/cmds/meson.py
@@ -716,6 +716,13 @@ def ipython(ctx, *, ipython_args, build_dir, pre_import=""):
         )
         ctx.invoke(build_cmd, build_dir=build_dir)
 
+    if not shutil.which("ipython"):
+        click.secho(
+            "Could not locate IPython; please install it using `pip install ipython` or similar.",
+            fg="bright_red",
+        )
+        sys.exit(1)
+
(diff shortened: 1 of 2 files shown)


oscarbenjamin:
Would it potentially make sense to change it here instead:
https://github.com/scientific-python/spin/blob/210ae8e894a855c6b68687179c14aebd28649659/spin/cmds/util.py#L67-L76

stefanv:
Ah, OK, I see we already do it for one branch, so then, yes, definitely.

stefanv:
OK, try now

PR #273, opened 2025-03-19
Add auto-completion docs

### LINKED ISSUE #269: Support tab auto-completion?
Hello! Thank you for writing a really nice package. It seems like exactly the type of thing I've been looking for.

Is it possible to implement tab autocompletion for the commands that are available to the developer? I'm not sure how much work this would be, but I think it would really improve the user experience of `spin`. I'm aware of how to do this directly in bash for things like Makefiles, but for the reasons listed in your README, I'm looking to move away from that workflow. Thanks for any help you can provide.

Edit: It seems like this might be a bit more involved than I originally thought, at least using the click shell completion: https://click.palletsprojects.com/en/stable/shell-completion/

stefanv:
Closes #269 

/cc @catanzaromj

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,7 +248,8 @@ cli = 'path/to/custom/meson'
 
 ## Auto-completion
 
-To enable shell auto-completion, first install `spin`, then follow these instructions.
+To enable shell auto-completion, first install `spin`, then follow these instructions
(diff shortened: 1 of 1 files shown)


catanzaromj:
I think it looks great. I might add a link to the click documentation on auto-completion for completeness. Thanks for the timely responses!

stefanv:
Thanks @catanzaromj!

PR #279, opened 2025-05-01
Expose `is_editable_install` and `editable_install_path`

stefanv:
Could be useful to downstream projects. See, e.g., https://github.com/scipy/scipy/issues/22887#issuecomment-2834063005


rgommers (approved):
LGTM, thanks!

PR #280, opened 2025-05-01
Improve traceback printout

stefanv:
Now, it prints the last frame of the traceback, in a different color from the message asking to report spin bugs if applicable.

See https://github.com/scipy/scipy/issues/22887#issuecomment-2834063005


rgommers:
Nice, thanks Stefan!

Issue #255, opened 2024-11-22
`spin lldb -- python -mspin test` doesn't work anymore


seberg:
I think this used to work, but I suspect that `python -mspin test` now runs in a new process so that the debugger/lldb doesn't attach to you.

I am not sure that this can be changed easily, FWIW, `python lldb -c ...`, etc. all work fine (I just like to run a test that fails if it does).

seberg:
The way spin spawns the actual spawns the test process, lldb doesn't wrap it.  Not sure if there is a clear solution though (maybe I need to tell lldb to wrap the child process or just launch the tests more explicitly).

Basically, add a test like:
```
def test_crash():
    import numpy as np
    from numpy.lib.stride_tricks import as_strided
    a = as_strided(np.arange(10), (10_000_000,), (8,))
    a[...] = 3
```
which segfaults.  If you just run it with `spin lldb -- python -mspin test -t ...` you should find that it's not running inside `lldb` and doesn't stop for debugging.  (Even disabling the fault handler with `PYTEST_ADDOPTS="-p no:faulthandler"`).

Maybe there is a very simple work-around.  I think last time I around I made a minimal example and used `spin lldb -c "exec(open('test.py').read())"` (to both run in the right process and to be sure to find the right `numpy` install).

stefanv:
I looked into this a bit, but I don't know lldb well enough to proceed.

I tried `-O settings set target.process.stop-on-exec false`, I tried to change `spin test` to replace the current process, etc. and I could not get lldb to see the traceback.

With the first option I get the traceback, but trying something like `bt` it says the process has aborted.

Are you able to help me explore the solutions here?

seberg:
Happy to poke this a bit, but not sure what to actually try.  I can't find a good hint on how one might tell `lldb` to follow the new process better.
I.e. `settings set target.process.stop-on-exec false` or `target.process.follow-fork-mode (parent|child)` has no effect?  ([This thread](https://github.com/vadimcn/codelldb/issues/317#issuecomment-636601305) makes me think maybe it just can't work on MacOS, but I am not sure it is the same problem.)

One thing to help if you poke more might be running with:
```
PYTEST_ADDOPTS=" -p no:faulthandler"
```
That will disable the Python traceback from the faulthandler.  But I don't recall it should change behavior, but better safe than sorry.

> I tried to change spin test to replace the current process, etc. and I could not get lldb to see the traceback.

[...]

PR #281, opened 2025-05-29
Allow wrapping test run in lldb

stefanv:
An experiment to see if we can wrap our test runs in lldb, instead of figuring out all the correct arguments to pass to lldb.

Closes https://github.com/scientific-python/spin/issues/255


ngoldbaum:
Just a drive-by comment that while you're at it you should probably add `--gdb` too for Linux folks...

PR #282, opened 2025-05-31
Fix duplication of docs in `extend_command`

czgdp1807:
`extend_command` duplicates doc strings of commands decorated with `extend_command`.  Whatever commands I have tried in SciPy, all have docs from the original `spin` function and the SciPy function of the command. Take a look at the following examples,

`spin shell --help` (in SciPy),

```zsh
(scipy-dev) 20:05:56:~/Quansight/scipy % spin shell --help
Usage: spin shell [OPTIONS] [SHELL_ARGS]...

  💻 Launch shell with PYTHONPATH set

  SHELL_ARGS are passed through directly to the shell, e.g.:

  spin shell -- -c 'echo $PYTHONPATH'

  Ensure that your shell init file (e.g., ~/.zshrc) does not override the PYTHONPATH.

  💻 Launch shell with PYTHONPATH set

  SHELL_ARGS are passed through directly to the shell, e.g.:

  spin shell -- -c 'echo $PYTHONPATH'

  Ensure that your shell init file (e.g., ~/.zshrc) does not override the PYTHONPATH.
```

`spin docs --help` (in SciPy)

```zsh
(scipy-dev) 20:06:38:~/Quansight/scipy % spin docs --help 
Usage: spin docs [OPTIONS] [SPHINX_TARGET]

  📖 Build Sphinx documentation

  By default, SPHINXOPTS="-W", raising errors on warnings. To build without raising on warnings:

    SPHINXOPTS="" spin docs

  To list all Sphinx targets:

    spin docs targets

  To build another Sphinx target:

    spin docs TARGET

  📖 Build Sphinx documentation

  By default, SPHINXOPTS="-W", raising errors on warnings. To build without raising on warnings:

    SPHINXOPTS="" spin docs

  To list all Sphinx targets:

    spin docs targets

  To build another Sphinx target:

    spin docs TARGET

  E.g., to build a zipfile of the html docs for distribution:

    spin docs dist
```

I think the problem is with the following line,

https://github.com/scientific-python/spin/blob/fee6abe42cea118c2e4577cce5e0ca034c55748e/spin/cmds/util.py#L175

If both, `my_cmd.help` and `user_func.__doc__` will be present then both will be combined. In my opinion, better idea would be to give priority to `user_func.__doc__` if it is present. Otherwise, fallback to `my_cmd.help`.


rgommers:
This actually seems to be covered already by the `doc` keyword to `extend_command`:

https://github.com/scientific-python/spin/blob/fee6abe42cea118c2e4577cce5e0ca034c55748e/spin/cmds/util.py#L128-L130

Did you try using that to get the desired docstring contents?

rgommers:
So according to that documentation, adding `doc=""` should already remove the `spin` docs, and only pick up whatever is in the docstring of the command in the package itself.

PR #286, opened 2025-08-06
Use sys.executable to run pip in the 'install' command.

WarrenWeckesser:
When I install python from a source tarball with the standard `./configure` script, there is no executable entry point called `pip` (it is called `pip3`).

This change modifies how `pip` is run in the `install()` function so that `sys.executable` is used to run the `pip` module.

### DIFF
--- a/spin/cmds/pip.py
+++ b/spin/cmds/pip.py
@@ -1,3 +1,5 @@
+import sys
+
 import click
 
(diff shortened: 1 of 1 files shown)


stefanv (approved):

stefanv:
Thanks, @WarrenWeckesser!

PR #297, opened 2025-10-01
Allow click 8.3.x series, excluding 8.3.0

stefanv:
Click 8.3.0 broke spin because Sentinel objects are now passed instead of defaults in some instances:

https://github.com/pallets/click/issues/3065

8.3.1 will include two workarounds:

https://github.com/pallets/click/issues/3065#issuecomment-3334011895

I suspect

https://github.com/pallets/click/pull/3068

will address our problem.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,7 @@ classifiers = [
   "Programming Language :: Python :: 3 :: Only",
 ]
 dependencies = [
-  "click>=8,<8.3",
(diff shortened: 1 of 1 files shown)


stefanv:
Leave this PR open until 8.3.1 is released, then re-run CI.

stefanv:
Tests passed here and on main, but failing over at https://github.com/scientific-python/spin/pull/304 with the sentinel error I expected to see here. Will revert for now.

PR #305, opened 2025-11-27
Add wheel command to pyproject-build commands

### LINKED ISSUE #301: Add wheel building for build?
Currently, Spin has the capability to produce sdists with build. It would be rather straightforward to add wheel building as well. While many projects rely on cibuildwheel due to its broad support, I think it would be useful to have an easy way of building wheels as well.

Are you open to PRs to that end?

zklaus:
This adds a command to the pyproject-build commands that builds a wheel similar to the existing `sdist` command.

Closes #301.

### DIFF
--- a/spin/tests/test_build_cmds.py
+++ b/spin/tests/test_build_cmds.py
@@ -118,7 +118,7 @@ def test_sdist(example_pkg):
 
 def test_wheel(example_pkg):
     spin("wheel")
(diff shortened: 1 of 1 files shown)


spin/tests/test_build_cmds.py
```diff
@@ -116,6 +116,10 @@ def test_sdist(example_pkg):
     spin("sdist")
 
 
+def test_wheel(example_pkg):
+    spin("wheel")
```
stefanv: Maybe should check for artifact?

